PR #1, opened 2021-10-14
Update README.md

cyanzhong:
I personally think "powered by" sounds like it relies on GitHub's C implementation, but it's in fact a complete Swift rewrite (excellent job 👍)

Feel free to close if you think otherwise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Swift `Markdown` is a Swift package for parsing, building, editing, and analyzing Markdown documents.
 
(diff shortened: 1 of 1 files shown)


0xTim:
It does rely on it. It uses https://github.com/apple/swift-cmark/tree/gfm under the hood which is a fork of the original cmark library make to work with SwiftPM with what looks like GitHub's commits cherry picked in

rsalesas:
Merged into main as 04d333b and tagged 0.18.0. GitHub closed this rather than marking it merged because the merge was pushed from a local `--no-ff` (to keep the fork's `Merge <branch>: <summary>` commit convention) and the branch was deleted in the same breath — the commits are on main.

PR #2, opened 2021-10-16
Fix MarkupTest's missing array code convention

GREENOVER:
## Summary

Found the missing code convention of the array in the MarkupTest code. 
So, add it to this.

## Checklist
- [x] Added tests


franklinsch (commented):
I see this file is generally using trailing commas for the last elements of array—thanks for spotting this and opening a PR! If we do want to be consistent with the trailing commas here,  it looks like some other places are missing it, so we should update those as well I'd say.

GREENOVER:
@franklinsch 
I reviewed all the files to find and add two parts that did not fit the array convention.
I hope it will be reflected for the perfection of the code.

stareta1202:
@swift-ci please test

PR #8, opened 2021-10-21
Test Swift CI support

shahmishal:


Kyle-Ye:
I believe it seems to be a bug on Swift Complier on Linux.

Also the 2021-10-05 snapshot of swift is working both on Linux and macOS.(For the success of building swift-markdown)

Bug created at bugs.swift.org/browse/SR-15362

PR #10, opened 2021-11-02
Rename main.swift to MarkdownCommand.swift and add @main

### LINKED ISSUE #78: Drop Swift 5.3 Support
Is there a timeline we can drop support for Swift 5.3?

Currently there are 2 Package.swift files which has a maintenance cost (Sometimes PR only updates one while missing updating the another one to make them in sync. And we have to make another PR to fix it)

Also the Swift 5.3 limits a lot of nice Swift syntax and features when fixing bug or developing new features.

eg.
- #10 need `@main` support which was introduced by Swift 5.4
- Sometimes we write closure and does not provide a return value which is OK locally since we are using Package@swift5.5 and it will be inferred in recent swift versions(Maybe introduced by Swift 5.4, 5.5 or 5.6🤔)
But writing such code will make the CI fail😢.
```swift
- xx = xx.map { element in
+ xx = xx.map { element -> String in
```

## Other Info
Some popular swift packages (In the server side) make a commitment to provide a 2 year support(If a version is introduced 2 years ago then we can drop it)  or 3 swift version support(When swift 5.7 is released and supported we can drop support for Swift 5.4).

I'd like to know what swift-version supporting strategy will the Swift-Docc project take. Thanks.

> Some context when updating min-swift version to Swift 5.3 https://github.com/apple/swift-markdown/pull/4

Kyle-Ye:
Bug/issue #, if applicable: Close #78

## Summary

- Rename main.swift to MarkdownCommand.swift to better reflect the file's content.
- Adopt the @main feature of Swift.

## Dependencies

It depends on min swift version to "5.4" to use executableTarget.
(When the time is appropriate/5.4 [released on April 26, 2021] is widely adopted, will convert it from draft PR into PR)

## Checklist


Kyle-Ye:
@swift-ci please test

Package.swift
```diff
@@ -1,10 +1,8 @@
-// swift-tools-version:5.3
-// In order to support users running on the latest Xcodes, please ensure that
-// Package@swift-5.5.swift is kept in sync with this file.
+// swift-tools-version:5.5
```
Kyle-Ye, line 1: I think it is safe for us to bump to use Swift 5.5 (Released on SEPTEMBER 20, 2021)

d-ronnqvist (approved):

Kyle-Ye:
@swift-ci please test

PR #17, opened 2021-11-09
Update swift-argument-parser

jakepetroules:


QuietMisdreavus:
@swift-ci Please test

QuietMisdreavus (approved):

QuietMisdreavus:
Can you also make sure to update [the Package.swift for Swift-DocC](https://github.com/apple/swift-docc/blob/main/Package.swift)? This will cause issues over there if not.

jakepetroules:
> Can you also make sure to update [the Package.swift for Swift-DocC](https://github.com/apple/swift-docc/blob/main/Package.swift)? This will cause issues over there if not.

Yep, see also apple/swift#40103 which is trying to do a coordinate update across all affected repos.

PR #4, opened 2021-10-17
Bump Swift version

Kyle-Ye:
## Summary

Bump the Swift version Package.swift use to 5.4. And add a new 5.5 Package.swift file

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Ran the `./bin/test` script and it succeeded

PR #21, opened 2021-12-05
Fix wrong Package desciption text

Kyle-Ye:
Bug/issue #, if applicable: 

## Summary

As suggested in #4 by @franklinsch, we should "adding a comment such as apple/swift-driver@main/Package.swift#L2"

It seems there is some accidental mistake on the text in that related PR. This is a fix to that mistake.

### DIFF
--- a/Package@swift-5.5.swift
+++ b/Package@swift-5.5.swift
@@ -1,5 +1,5 @@
 // swift-tools-version:5.5
-// In order to support users running on legacy Xcodes, please ensure that
+// In order to support users running on previous versions of Xcode, please ensure that
 // Package.swift is kept in sync with this file.
 /*
  This source file is part of the Swift.org open source project
--- a/Package@swift-5.5.swift
+++ b/Package@swift-5.5.swift
@@ -1,6 +1,6 @@
 // swift-tools-version:5.5
-// In order to support users running on the latest Xcodes, please ensure that
-// Package@swift-5.5.swift is kept in sync with this file.
+// In order to support users running on legacy Xcodes, please ensure that
+// Package.swift is kept in sync with this file.
 /*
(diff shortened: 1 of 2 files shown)


Kyle-Ye:
@swift-ci please test

Package@swift-5.5.swift
```diff
@@ -1,6 +1,6 @@
 // swift-tools-version:5.5
-// In order to support users running on the latest Xcodes, please ensure that
-// Package@swift-5.5.swift is kept in sync with this file.
+// In order to support users running on legacy Xcodes, please ensure that
```
franklinsch, line 2: Nit: I think "legacy" might not always be appropriate term to use for clients of the `Package.swift` file. I suggest relaxing the phrasing a bit:

```suggestion
// In order to support users running on previous versions of Xcode, please ensure that
```

franklinsch (commented):
Thanks for opening this! I just have a small suggestion for the phrasing.

Package@swift-5.5.swift
Kyle-Ye: https://github.com/apple/swift-driver/blob/9ef905a68300e82e4dd72fab2ecf369a917b09b1/Package%40swift-5.5.swift#L1-L3

The swift-driver uses "legacy" here. But if you prefer, we can change it.

franklinsch: I think "previous versions of Xcode" is more accurate because I think we want to be able to copy/paste this file for future versions of Swift compilers supported by Xcode versions in beta. In those cases, "legacy" wouldn't be appropriate to refer to the version of Xcode that's shipping in the Mac App Store.

Kyle-Ye:
@swift-ci Please test

franklinsch (approved):

PR #22, opened 2021-12-08
store ordered list start index from cmark

QuietMisdreavus:
Bug/issue #, if applicable: rdar://73847907

## Summary

While cmark-gfm stores the start index for an ordered list, swift-markdown currently discards this information. This PR is part of a set to use this start index when handling ordered lists.

I also added license headers to the markdown in the doc bundle so that `bin/check-source` can pass.

## Dependencies

None

## Testing

With the following markdown test file, ensure that the debug tree properly contains the start index of the list.

```markdown
2. this is a test
3. this is a test
4. this is a test
```

Steps:
1. `swift run markdown-tool dump-tree test.md`
2. Compare the output to the below sample and ensure that the start index is preserved:

```
Document
└─ OrderedList start: 2
   ├─ ListItem
   │  └─ Paragraph
   │     └─ Text "this is a test"
   ├─ ListItem
   │  └─ Paragraph
   │     └─ Text "this is a test"
   └─ ListItem
      └─ Paragraph
         └─ Text "this is a test"
```

## Checklist

- [x] Added tests


QuietMisdreavus:
@swift-ci Please test

Sources/Markdown/Block Nodes/Block Container Blocks/OrderedList.swift
```diff
@@ -29,7 +29,26 @@ public extension OrderedList {
     // MARK: ListItemContainer
 
     init<Items: Sequence>(_ items: Items) where Items.Element == ListItem {
-        try! self.init(.orderedList(parsedRange: nil, items.map { $0.raw.markup }))
+        try! self.init(.orderedList(parsedRange: nil, items.map { $0.raw.markup }, start: nil))
+    }
+
+    /// The starting index for the list.
+    ///
+    /// If this is `nil`, the list will start at the default value of 1.
```
franklinsch, line 37: I think the phrasing could be improved here, I think what we mean is that if `start` is `nil`, it's up to the client to pick a start index? (1, most likely)

Sources/Markdown/Parser/CommonMarkConverter.swift
```diff
@@ -299,7 +299,9 @@ struct MarkupParser {
         case CMARK_BULLET_LIST:
             return MarkupConversion(state: childConversion.state.next(), result: .unorderedList(parsedRange: parsedRange, childConversion.result))
         case CMARK_ORDERED_LIST:
-            return MarkupConversion(state: childConversion.state.next(), result: .orderedList(parsedRange: parsedRange, childConversion.result))
+            let cmarkStart = cmark_node_get_list_start(state.node)
+            let start: Int? = cmarkStart == 1 ? nil : Int(cmarkStart)
```
franklinsch, line 303: Nit: Alternatively, we could make `start` non-optional and allow storing `1`. Then, clients would hardcode any specific behavior they want when start=1 (in DocC's case, to not emit the `start` property in render node).

Sources/Markdown/Walker/Walkers/MarkupFormatter.swift
```diff
@@ -492,15 +492,23 @@ public struct MarkupFormatter: MarkupWalker {
...
         }
         let numeral: UInt
         switch formattingOptions.orderedListNumerals {
         case let .allSame(n):
-            numeral = n
+            if let origStart = list.start {
+                numeral = UInt(origStart)
+            } else {
+                numeral = n
+            }
```
franklinsch, line 505: Nit: 

```suggestion
            numeral = list.start.map(UInt.init) ?? n
```

same below

Sources/Markdown/Walker/Walkers/MarkupTreeDumper.swift
```diff
@@ -189,6 +189,14 @@ struct MarkupTreeDumper: MarkupWalker {
         dump(heading, customDescription: "level: \(heading.level)")
     }
 
+    mutating func visitOrderedList(_ orderedList: OrderedList) {
```
franklinsch, line 192: Huh, was this just missing previously?

Sources/Markdown/Base/RawMarkup.swift
```diff
@@ -24,7 +24,7 @@ enum RawMarkupData: Equatable {
     case thematicBreak
     case htmlBlock(String)
     case listItem(checkbox: Checkbox?)
-    case orderedList
+    case orderedList(start: Int?)
```
franklinsch, line 27: Since this is an API breaking change, this will require a coordinated change with Swift-DocC. Is there a way we could introduce the change without requiring the coordination?

franklinsch: Also, I think this could benefit from a default value (`orderedList(start: Int? = nil)`)

Sources/Markdown/Walker/Walkers/MarkupTreeDumper.swift
QuietMisdreavus: It was using the default visitor before, since it didn't have any associated data to print.

Sources/Markdown/Base/RawMarkup.swift
QuietMisdreavus: The `RawMarkup` type isn't used in Swift-DocC, so this won't cause source breakage there. The public accessor to the start index is the new `start` property on the `OrderedList` type.

QuietMisdreavus: Looking at this again, this type (`RawMarkupData`) isn't public API. Other enum cases don't use a default value; that's enforced via the public `Markup` types which wrap `RawMarkup`. I don't think it would be useful to include a default value for the list start value here; that comes from cmark.

QuietMisdreavus:
I've rebased this onto the latest `main` (and dropped the commit that added copyrights to the doc markdown files, since that was added separately). I've also updated the doc comment for `OrderedList.start` to better reflect the default value, and updated the implementation to allow ordered lists to start with zero (since cmark and HTML allow this).

Sources/Markdown/Base/RawMarkup.swift
```diff
@@ -24,7 +24,7 @@ enum RawMarkupData: Equatable {
     case thematicBreak
     case htmlBlock(String)
     case listItem(checkbox: Checkbox?)
-    case orderedList
+    case orderedList(start: UInt = 1)
```
franklinsch, line 27: thoughts on calling this `startIndex` throughout to make it clearer what 'start' we're referring to?

Sources/Markdown/Walker/Walkers/MarkupFormatter.swift
```diff
@@ -492,15 +492,23 @@ public struct MarkupFormatter: MarkupWalker {
...
             return nil
         }
         let numeral: UInt
         switch formattingOptions.orderedListNumerals {
         case let .allSame(n):
-            numeral = n
+            if list.start == 1 {
+                numeral = n
+            } else {
+                numeral = UInt(list.start)
```
franklinsch, line 504: Is this right? Should we not preserve the numbering provided to `allSame`?

Sources/Markdown/Walker/Walkers/MarkupFormatter.swift
```diff
@@ -492,15 +492,23 @@ public struct MarkupFormatter: MarkupWalker {
...
+            }
         case let .incrementing(start):
-            numeral = start + UInt(listItem.indexInParent)
+            if list.start != 1 {
+                numeral = UInt(list.start) + UInt(listItem.indexInParent)
```
franklinsch, line 508: Same question here

Sources/Markdown/Walker/Walkers/MarkupFormatter.swift
QuietMisdreavus: Looking back at this, the existing implementation should defer to the indices given to `allSame` and `incrementing`. However, now that ordered lists can store their own starting index, we should allow this API to defer to that. I filed https://github.com/apple/swift-markdown/issues/76 to track that. That can be taken care of in a separate PR, or handled here.

Sources/Markdown/Base/RawMarkup.swift
QuietMisdreavus: Sounds good, i've pushed this change up.

QuietMisdreavus:
@swift-ci Please test

QuietMisdreavus:
Had to update a test after reverting the change to the markup formatter.

@swift-ci Please test

QuietMisdreavus:
@swift-ci Please test

QuietMisdreavus:
@swift-ci Please test

PR #34, opened 2022-04-05
Enable library evolution via a conditional

franklinsch:
Bug/issue #, if applicable: rdar://91237377

## Summary

Adds support for building the `Markdown` library with library evolution enabled. This is disabled by default, and to enable the behavior, add an `.enable-library-evolution` file at the root of your checkout.

## Dependencies

None.

## Testing

Verify that `Markdown` is built with library evolution enabled by inspecting build logs if and only if the checkout has an `.enable-library-evolution` file at its root.

## Checklist

- ~[ ] Added tests~ N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@
 /*.xcodeproj
 .swiftpm
 Package.resolved
+.enable-library-evolution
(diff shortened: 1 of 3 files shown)


franklinsch:
@swift-ci please test

franklinsch:
This change didn't end up suiting my needs because I couldn't find a reliable way to always generate a `.swiftinterface`, so I will close this PR for now.

PR #36, opened 2022-05-07
Improve performance of `Markup.child(at:)`

ethan-kusters:
## Summary

While investigating general performance improvements in Swift-DocC for:

- apple/swift-docc#166

I found that Swift-DocC was spending a lot of time creating iterators while rewriting markup elements in Swift-Markdown code. This PR is a refactoring of the `Markup.child(at:)` method to improve performance from `O(childCount)` to `O(1)`.

The change here is to just pull in the code from [`MarkupChildren.Iterator`](https://github.com/apple/swift-markdown/blob/2ba1ed5/Sources/Markdown/Base/MarkupChildren.swift#L26) that was already being used:

```swift
public mutating func next() -> Markup? {
    let index = childMetadata.indexInParent
    guard index < parent.childCount else {
        return nil
    }
    let rawChild = parent.raw.markup.child(at: index)
    let absoluteRawChild = AbsoluteRawMarkup(markup: rawChild, metadata: childMetadata)
    let data = _MarkupData(absoluteRawChild, parent: parent)
    childMetadata = childMetadata.nextSibling(from: rawChild)
    return makeMarkup(data)
}
```

and call `RawMarkup.child(at:)` directly in the `Markup.child(at:)` method instead of wrapping those calls in the creation of the markup iterator.

## Testing

Existing tests that exercise `Markup.child(at:)` continue to pass as does the Swift-DocC test suite when using this change.

Since this is a 1-1 refactoring I didn't see an opportunity for new tests but happy to add any that would be helpful here.

## Checklist

- [ ] Added tests

### DIFF
--- a/Sources/Markdown/Base/Markup.swift
+++ b/Sources/Markdown/Base/Markup.swift
@@ -183,14 +183,16 @@ extension Markup {
 
     /// Returns the child at the given position if it is within the bounds of `children.indices`.
     ///
-    /// - Complexity: `O(childCount)`
+    /// - Complexity: `O(1)`
     public func child(at position: Int) -> Markup? {
         precondition(position >= 0, "Cannot retrieve a child at negative index: \(position)")
-        guard position <= raw.markup.childCount else {
+        guard position < raw.markup.childCount else {
             return nil
         }
-        var iterator = children.dropFirst(position).makeIterator()
-        return iterator.next()
+        let rawChild = raw.markup.child(at: position)
+        let absoluteRawMarkup = AbsoluteRawMarkup(markup: rawChild, metadata: raw.metadata.firstChild())
+        let data = _MarkupData(absoluteRawMarkup, parent: self)
+        return makeMarkup(data)
     }
 
     /// Traverse this markup tree by descending into the child at the index of each path element, returning `nil` if there is no child at that index or if the expected type for that path element doesn't match.


ethan-kusters:
@swift-ci please test

Sources/Markdown/Base/Markup.swift
```diff
@@ -183,14 +183,16 @@ extension Markup {
...
     public func child(at position: Int) -> Markup? {
         precondition(position >= 0, "Cannot retrieve a child at negative index: \(position)")
-        guard position <= raw.markup.childCount else {
+        guard position < raw.markup.childCount else {
             return nil
         }
-        var iterator = children.dropFirst(position).makeIterator()
-        return iterator.next()
+        let rawChild = raw.markup.child(at: position)
+        let absoluteRawMarkup = AbsoluteRawMarkup(markup: rawChild, metadata: raw.metadata.firstChild())
```
bitjammer, line 193: This is not correct. You’re using the first child’s metadata for the nth child.

bitjammer (changes requested):
Sorry, but this is not a sound change. In order to maintain absolute identity of nodes without using class pointers, identifiers are calculated on-the-fly as you traverse the tree, so we can't just skip ahead anyway. Is the time spent initializing iterators or simply traversing the tree?

There really shouldn't be a lot of `child(at:)` calls in rewriters. For example, are they mostly `child(at: 0)` or `child(at: 1)`? I recommend we schedule some time to talk more about the performance analysis and take a look at the use sites that you think are problematic. There may be some other things that we can in DocC as well. Until we do that, I don't think we should make a change like this.

ethan-kusters:
> Sorry, but this is not a sound change.

I thought I might be missing something here- thanks for the review 😄 

I'll set up some some time for us to walk through this.

Sources/Markdown/Base/Markup.swift
ethan-kusters: Is this incorrect in the iterator then too or am I misunderstanding something here?

https://github.com/apple/swift-markdown/blob/2ba1ed5/Sources/Markdown/Base/MarkupChildren.swift#L23

bitjammer:
No worries, anytime! And thank you for looking into it.

PR #44, opened 2022-05-14
Improve performance of `Markup.child(at:)` method while still accurately tracking metadata

ethan-kusters:
## Summary

While investigating general performance improvements in Swift-DocC for:

- https://github.com/apple/swift-docc/pull/166

I found that Swift-DocC was spending a lot of time creating iterators while rewriting markup elements in Swift-Markdown code using `MarkupRewriter`s. This PR is a refactoring of the `Markup.child(at:)` method to improve performance.

I put up a PR with a similar change in #36 where @bitjammer pointed out that I was incorrectly tracking metadata for child elements. This is a second iteration on that PR that resolves that issue and adds a test to catch the issue I missed before. It appears to have the same performance win as the original PR when testing with the Swift-DocC performance suite.

## Testing

I added tests that confirm the metadata returned for child elements via the new `child(at:)` implementation is the same as the metadata for child elements returned via the existing sequence implementation. Swift-Markdown and Swift-DocC's existing tests continue to pass with this change.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran the `./bin/test` script and it succeeded
- [x] Updated documentation if necessary

PR #45, opened 2022-05-21
Correctly set `indexInParent` in `Markup.child(at:)`

ethan-kusters:
## Summary

This fixes a regression introduced with #44 where `Markup.child(at:)` began returning markup with incorrect metadata resulting in out-of-bounds errors.

## Testing

Existing and new unit tests pass. I've also confirmed that Swift-DocC's test suite is now passing with this change (it fails with what's currently on `main`).

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/Markdown/Base/Markup.swift
+++ b/Sources/Markdown/Base/Markup.swift
@@ -204,7 +204,7 @@ extension Markup {
                 childId: firstChildID.childId + siblingSubtreeCount
             )
             
-            childMetadata = MarkupMetadata(id: childID, indexInParent: indexInParent + position)
+            childMetadata = MarkupMetadata(id: childID, indexInParent: position)
         }
         
         let rawChild = raw.markup.child(at: position)
--- a/Tests/MarkdownTests/Base/MarkupTests.swift
+++ b/Tests/MarkdownTests/Base/MarkupTests.swift
@@ -348,6 +348,25 @@ final class MarkupTests: XCTestCase {
         }
     }
     
+    func testNestedChildAtPositionHasCorrectMetadata() throws {
+        let source = "This is some **bold** and *italic* text and a [multi-formatted *link* **to** `github`](github.com)."
+        
+        let document = Document(parsing: source)
+        let link = try XCTUnwrap(document.child(at: 0)?.child(at: 5) as? Link)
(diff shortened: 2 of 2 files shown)


Sources/Markdown/Base/Markup.swift
```diff
@@ -204,7 +204,7 @@ extension Markup {
                 childId: firstChildID.childId + siblingSubtreeCount
             )
             
-            childMetadata = MarkupMetadata(id: childID, indexInParent: indexInParent + position)
+            childMetadata = MarkupMetadata(id: childID, indexInParent: position)
```
ethan-kusters, line 207: The `indexInParent` value of a child is unrelated to the parent's `indexInParent`. I missed this initially because tests were only checking for children of the first item where `indexInParent` would be 0. This PR includes a new test that asserts `Markup.child(at:)` returns correct values for children of nested items as well.

ethan-kusters:
@swift-ci please test

Tests/MarkdownTests/Base/MarkupTests.swift
```diff
@@ -348,6 +348,25 @@ final class MarkupTests: XCTestCase {
...
+        
+        for (index, sequencedChild) in link.children.enumerated() {
+            let indexedChild = try XCTUnwrap(link.child(at: index))
+            
+            let indexedChildMetadata = indexedChild.raw.metadata
+            let sequencedChildMetadata = sequencedChild.raw.metadata
+            
+            XCTAssertEqual(indexedChildMetadata.id, sequencedChildMetadata.id)
+            XCTAssertEqual(indexedChildMetadata.indexInParent, sequencedChildMetadata.indexInParent)
+            XCTAssertEqual(indexedChildMetadata.indexInParent, index)
```
ethan-kusters, line 366: These assertions failed with the implementation initially landed in #44.

bitjammer: 👍

Sources/Markdown/Base/Markup.swift
bitjammer: Heh, nice catch. Sorry I also missed that.

bitjammer (approved):

PR #53, opened 2022-06-06
Add simple playground support

### LINKED ISSUE #52: Add Snippets playground support
Add Playground or Snippets Support, so that people can quickly play with Markdown API

Kyle-Ye:
Bug/issue #, if applicable: 

Close #52 

## Summary

Add simple playground support

## Testing

image

## Checklist

- [x] Added tests

### DIFF
--- a/Playgrounds/Markdown.playground/Pages/Overview.xcplaygroundpage/Contents.swift
+++ b/Playgrounds/Markdown.playground/Pages/Overview.xcplaygroundpage/Contents.swift
@@ -1,0 +1,17 @@
+/*:
+ # Markdown
+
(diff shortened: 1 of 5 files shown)


franklinsch:
Per https://github.com/apple/swift-markdown/issues/52, I'd much rather this uses Snippets rather than Playgrounds, so that the content can be used on all platforms :)

Kyle-Ye:
One idea is we add Xcode Playground first. And when Snippets are landed, we change the playground of mark-down from `Xcode Playground(macOS only)` form to `Snippets(Cross-platform)` form.

It is really handy to test some data when implementation something. Otherwise, we have to new a Package to dependent on swift-markdown or add something temporary code on MarkdownTests target.

Issue #59, opened 2022-06-13
Support for custom aside titles


franklinsch:
#### Description:

Swift-Markdown supports authoring 'aside' (sometimes called 'callout') content using the following syntax:

```md
> Important: This is important!
```

However, it's not currently possible to use a custom title for an aside. The ones supported by Swift-Markdown are listed here: https://github.com/apple/swift-markdown/blob/main/Sources/Markdown/Interpretive%20Nodes/Aside.swift#L27.

#### Motivation:

Authors should have the flexibility to customize the title of their aside sections, for specific use-cases they might have. Syntax is TBD, but the following might work:

```md
> My Custom Title: This is an aside.
```

#### Importance:

This feature is an incremental improvement over the existing support for asides.

PR #60, opened 2022-06-15
Update Aside to parse custom titles

micpringle:
Bug/issue #, if applicable: rdar://65856948 #59

## Summary

Adds support for parsing custom aside titles.

> **Note**: This is a source-breaking change for the `Aside.Kind` API and removes case-insensitivity support for aside tags. 

## Dependencies

https://github.com/apple/swift-docc/pull/303

## Testing

See https://github.com/apple/swift-docc/pull/303 for details.

## Checklist

- [X] Added tests
- [X] Updated documentation if necessary

### DIFF
--- a/Sources/Markdown/Interpretive Nodes/Aside.swift
+++ b/Sources/Markdown/Interpretive Nodes/Aside.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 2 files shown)


micpringle:
@swift-ci please test

Sources/Markdown/Interpretive Nodes/Aside.swift
```diff
@@ -21,87 +21,155 @@ import Foundation
 /// > different meaning, as it doesn't quote speech.
 /// ```
 public struct Aside {
-    /// The kind of aside.
-    public enum Kind: String, CaseIterable {
+    /// Describes the different kinds of aside.
+    public struct Kind: CaseIterable, Equatable {
```
franklinsch, line 25: This is an API breaking change but it makes sense to me, `Kind` shouldn't be an enum because it no longer represents a closed set. Can we call this out in the PR description?

franklinsch (approved):

Sources/Markdown/Interpretive Nodes/Aside.swift
```diff
@@ -21,87 +21,155 @@ import Foundation
...
+        
         /// A "version" aside.
-        case version = "Version"
-
+        public static let version = Kind(displayName: "Version")
+        
         /// A "throws" aside.
-        case `throws` = "Throws"
-
-        public init?(rawValue: String) {
```
franklinsch, line 95: Actually, I think we should keep this initializer and mark it as deprecated so that clients have time to adopt `init(displayName:)` instead.

Kyle-Ye (commented):
A small suggestion: Could we update the year number from "2021" to "2021-2022"?

Kyle-Ye:
I agree we should use `struct` instead of `enum` to make it possible for client to create custom aside.

But I think we could keep rawValue property and make `Aside` conforms to `RawRepresentable`.(Since IMO Swift prefers using protocol to constraint behavour) If we think `displayName ` is a better name for client usage, we can add the following:
```swift
var displayName: String {
   rawValue // Or some logic on rawValue
}
```
And add warning or deprecate info on `rawValue` property.

micpringle:
@swift-ci please test

Sources/Markdown/Interpretive Nodes/Aside.swift
micpringle: Great suggestion @Kyle-Ye. I've pushed changes that take this approach instead.

micpringle:
@Kyle-Ye I've incorporated all of your suggestions. Please take another look.

Sources/Markdown/Interpretive Nodes/Aside.swift
```diff
@@ -21,87 +21,116 @@ import Foundation
...
+                `throws`,
+            ]
+        }
+        
+        /// The underlying raw string value.
+        public var rawValue: String
+        
+        /// Creates an aside kind with the specified raw value.
+        /// - Parameter rawValue: The string the aside displays as its title.
         public init?(rawValue: String) {
```
Kyle-Ye, line 129:
```suggestion
        public init(rawValue: String) {
```

Kyle-Ye: Since the initializer never fails. We should use a non-failable initializer and update all the `Aside(rawValue: xx)!` to `Aside(rawValue: xx)`

Sources/Markdown/Interpretive Nodes/Aside.swift
```diff
@@ -116,14 +145,13 @@ public struct Aside {
...
-        let firstColonIndex = initialText.string.firstIndex(where: { $0 == ":" }),
-        let kind = Kind(rawValue: String(initialText.string[initialText.string.startIndex..<firstColonIndex])) else {
+        let firstColonIndex = initialText.string.firstIndex(where: { $0 == ":" }) else {
             // Otherwise, default to a note aside.
             self.kind = .note
             self.content = Array(blockQuote.blockChildren)
             return
         }
-        self.kind = kind
+        self.kind = Kind(rawValue: String(initialText.string[initialText.string.startIndex..<firstColonIndex]))!
```
Kyle-Ye, line 154:
```suggestion
        self.kind = Kind(rawValue: String(initialText.string[..<firstColonIndex]))!
```

micpringle: Fixed.

micpringle:
@swift-ci please test

Sources/Markdown/Interpretive Nodes/Aside.swift
micpringle: Hey @Kyle-Ye. 👋 

The initializer needs to be failable to meet the requirements of `RawRepresentable`. By keeping it failable, it also means existing client code will continue to work instead of introducing additional breaking changes.

Right now the initializer never fails, but in the future we may want to change this. For  example, returning `nil` if `rawValue` is empty or contains only whitespace.

franklinsch: +1 on keeping the type `RawRepresentable` and hence, making the initializer failable. This type is essentially a wrapper for raw string values, so `RawRepresentable` makes sense.

franklinsch (approved):

Sources/Markdown/Interpretive Nodes/Aside.swift
Kyle-Ye: > Hey @Kyle-Ye. 👋
> 
> The initializer needs to be failable to meet the requirements of `RawRepresentable`. By keeping it failable, it also means existing client code will continue to work instead of introducing additional breaking changes as we move from an `enum` to a `struct`.
> 
> Right now the initializer never fails, but in the future we may want to change this. For example, returning `nil` if `rawValue` is empty or contains only whitespace.

1. We can use a non-failable init to meet failable init requirement. That's totally OK in swift.
2. But considering the original enum's init is `init?(rawValue: String)` and we may return `nil` in the future. I agree to keep the failable init behavior.

Kyle-Ye (approved):

micpringle:
@swift-ci please test

PR #66, opened 2022-08-01
Fix single-line directive parsing issue

### LINKED ISSUE #64: Including a @Comment in silently breaks symbol's doc rendering
### Description

Let's say I have a symbol `Example`.

In my `Sources/ExampleThing/ExampleThing.docc/Example.md` i have

```md
# ``Example``

@Comment { Don't forget X }

## More things here

Lorem ipsum...
```

This silently breaks the rendering of the entire page's rendering -- it appears completely empty and no warnings are emitted either.

I'm guessing `@Comment` can only be used in a tutorial -- which is a bit meh... the syntax should be the same in al types of files.

### Checklist

- [ ] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/apple/swift-docc/issues).

### Expected Behavior

The @Comment should work as expected -- just not be rendered.

### Actual behavior

The rendering of the entire documentation page for the symbol is empty / broken by the presence of a `@Comment`

### Steps To Reproduce

- document any symbol
- add `@Comment { hello }`
- observe the comment prevented content, other than the sub-title from being rendered

### Swift-DocC Version Information

5.7

### Swift Compiler Version Information

```shell
-> % swift --version
swift-driver version: 1.62.1 Apple Swift version 5.7 (swiftlang-5.7.0.123.6 clang-1400.0.29.50)
```

### LINKED ISSUE #65: DocC drops content have single-line directive
### Description

When compiling this documentation comment:

```swift
/// Hello
///
/// Some content
///
/// @Comment { This is a comment }
///
/// Hello World
```

The text after the comment directive incorrectly gets dropped. This doesn't reproduce when writing the directive over multiple lines.

### Checklist

- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/apple/swift-docc/issues).

### Expected Behavior

DocC should not drop content after single-line directives.

### Actual behavior

_No response_

### Steps To Reproduce

_No response_

### Swift-DocC Version Information

Swift 5.7

### Swift Compiler Version Information

_No response_

Kyle-Ye:
Bug/issue #, if applicable: 

- [ ] Close #64
- [x] Close #65

## Summary

Before this PR, the block directive parser will just ignore the content after "{" in the same line.

After the PR, the behavior becomes the following:
```
@xx { yy           // "yy" will be ignored
@xx { yy }         // "yy" will be parsed
@xx { yy } zz }    // "yy } zz" will be parsed. It can be "yy } zz" or "yy". My implementation choose it to the former.
```
## Checklist

- [x] Added tests

### DIFF
--- a/Tests/MarkdownTests/Parsing/BlockDirectiveParserTests.swift
+++ b/Tests/MarkdownTests/Parsing/BlockDirectiveParserTests.swift
@@ -955,4 +955,31 @@ class BlockDirectiveArgumentParserTests: XCTestCase {
         """#
         XCTAssertEqual(document.debugDescription(options: .printSourceLocations), expectedDump)
     }
+    
+    func testSingleLineDirective() {
+        let source = """
+        @xx { yy }
+        @xx { yy } zz }
+        @xx { yy
+            z
+        }
+        Hello
+        """
+        let document = Document(parsing: source, options: [.parseBlockDirectives])
+        let expectedDump = #"""
+        Document @1:1-6:6
+        ├─ BlockDirective @1:1-1:11 name: "xx"
+        │  └─ Paragraph @1:7-1:9
+        │     └─ Text @1:7-1:9 "yy"
+        ├─ BlockDirective @2:1-2:16 name: "xx"
+        │  └─ Paragraph @2:7-2:14
+        │     └─ Text @2:7-2:14 "yy } zz"
+        ├─ BlockDirective @3:1-5:2 name: "xx"
+        │  └─ Paragraph @4:5-4:6
+        │     └─ Text @4:5-4:6 "z"
+        └─ Paragraph @6:1-6:6
+           └─ Text @6:1-6:6 "Hello"
+        """#
+        XCTAssertEqual(document.debugDescription(options: .printSourceLocations), expectedDump)
+    }
 }
--- a/Sources/Markdown/Parser/BlockDirectiveParser.swift
+++ b/Sources/Markdown/Parser/BlockDirectiveParser.swift
@@ -69,6 +69,9 @@ struct PendingBlockDirective {
     /// The location of the last character accepted into the block directive.
     var endLocation: SourceLocation
 
+    /// The pending line of the block directive
+    var pendingLine: TrimmedLine?
+
     /// `true` if the block directive container is expecting child content,
     /// i.e. it has parsed an opening curly brace `{`.
     var isAwaitingChildContent: Bool {
@@ -164,7 +167,7 @@ struct PendingBlockDirective {
 
     /// Continue parsing from the `contentsEnd` state.
     @discardableResult
-    mutating func parseContentsEnd(from line: TrimmedLine) -> Bool  {
+    mutating func parseContentsEnd(from line: TrimmedLine) -> Bool {
         precondition(isAwaitingChildContent)
         var line = line
         line.lexWhitespace()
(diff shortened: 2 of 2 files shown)


Kyle-Ye:
@swift-ci please test

Kyle-Ye:
Ping @QuietMisdreavus . Could you help review this?

Tests/MarkdownTests/Parsing/BlockDirectiveParserTests.swift
```diff
@@ -955,4 +955,31 @@ class BlockDirectiveArgumentParserTests: XCTestCase {
         """#
         XCTAssertEqual(document.debugDescription(options: .printSourceLocations), expectedDump)
     }
+    
+    func testSingleLineDirective() {
+        let source = """
+        @xx { yy }
+        @xx { yy } zz }
+        @xx { yy
```
QuietMisdreavus, line 963: Was there a specific reason to drop this trailing content? Part of me thinks it should be included in the directive body.

Kyle-Ye: That's interesting. But it may need more work to achieve this.
And I think this kind of usage is very rare and is not recommended.
We should use either
```swift
@xx {
	yy
	z
}
```
or
```swift
@xx { yy z }
```

Since this PR is an additive and is an implementation-only detail.

I think this is room for future PR to support this if you think this kind of grammar should be supported.

QuietMisdreavus: I agree that enabling the `@xx { yy z }` form is more important than getting everything right immediately. This looks good to go.

QuietMisdreavus (approved):
Thanks so much!

Kyle-Ye:
#64 and #65 seem to be different issue for me. This PR only solved #65, #64 is still unresolved. cc @QuietMisdreavus

QuietMisdreavus:
@Kyle-Ye They seem to be identical to me - a block directive that has its closing brace on the same line as its name treats the rest of the document as part of the directive. When that directive is `@Comment`, Swift-DocC treats the entire rest of the document as a comment, and removes it from the rendering. In #64, the comment was added directly underneath the top-level heading, which then swallowed the entire page's content, resulting in an empty page. In #65, the comment was farther down the page, and swallowed only the content between it and the end of the document. In both cases, the directive was never closed in the parser, so everything written after it was treated as part of the directive. The only difference between the two issues (that i can see) is where in the file they were written.

Kyle-Ye:
> @Kyle-Ye They seem to be identical to me - a block directive that has its closing brace on the same line as its name treats the rest of the document as part of the directive. When that directive is `@Comment`, Swift-DocC treats the entire rest of the document as a comment, and removes it from the rendering. In #64, the comment was added directly underneath the top-level heading, which then swallowed the entire page's content, resulting in an empty page. In #65, the comment was farther down the page, and swallowed only the content between it and the end of the document. In both cases, the directive was never closed in the parser, so everything written after it was treated as part of the directive. The only difference between the two issues (that i can see) is where in the file they were written.

Yes, I agree that they are identical issues from the perspective of markdown. But for some undiscovers other reasons, even if we fix the parsing issue of `@Comment` (and other block directive), #65 still remains unresolved.

1. I think we could merge this and close #64. And later open another PR whether on swift-docc or swift-markdown to fix it.
2. Or we should delay merging this until we find a solution for #65.
What do you think of it? @QuietMisdreavus

QuietMisdreavus:
Forgive my single-mindedness, but i just don't see what you mean by #65 not being fixed. What's the remaining issue?

Kyle-Ye:
> Forgive my single-mindedness, but i just don't see what you mean by #65 not being fixed. What's the remaining issue?

Sorry I mixed the order of the 2 issues. #65 is confirmed to to fixed while #64 is not.

QuietMisdreavus:
@Kyle-Ye With this PR and Swift-DocC `113ffa3` (not `main` because this branch would need to rebase on Swift-Markdown's `main` first), they all render identically to me:

image

image

image

The only thing i changed with the samples you added to the other issue was to make them standalone articles so i could render them side-by-side. Their Render JSON is identical except for the title and URL. The issue with syntax highlighting is specifically an issue with Xcode and needs to be handled separately (the feedback you filed is where that will be worked on). Is there something i'm still missing?

Kyle-Ye:
@swift-ci please test

PR #72, opened 2022-08-28
Update README.md

mikekreuzer:
Update example Package.swift syntax

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,13 +11,13 @@ The markup tree provided by this package is comprised of immutable/persistent, t
 In your `Package.swift` Swift Package Manager manifest, add the following dependency to your `dependencies` argument:
 
 ```swift
-.package(url: "https://github.com/apple/swift-markdown.git", .branch("main")),
+.package(url: "https://github.com/apple/swift-markdown.git", branch: "main"),
 ```
 
 Add the dependency to any targets you've declared in your manifest:
 
 ```swift
-.target(name: "MyTarget", dependencies: ["Markdown"]),
+.product(name: "Markdown", package: "swift-markdown"),
 ```
 
 To parse a document, use `Document(parsing:)`, supplying a `String` or `URL`:
--- a/README.md
+++ b/README.md
@@ -17,7 +17,9 @@ In your `Package.swift` Swift Package Manager manifest, add the following depend
 Add the dependency to any targets you've declared in your manifest:
 
 ```swift
-.product(name: "Markdown", package: "swift-markdown"),
+.target(name: "MyTarget", dependencies: [
+    .product(name: "Markdown", package: "swift-markdown"),
+]),
 ```
 
 To parse a document, use `Document(parsing:)`, supplying a `String` or `URL`:


README.md
```diff
@@ -11,13 +11,13 @@ The markup tree provided by this package is comprised of immutable/persistent, t
 In your `Package.swift` Swift Package Manager manifest, add the following dependency to your `dependencies` argument:
 
 ```swift
-.package(url: "https://github.com/apple/swift-markdown.git", .branch("main")),
+.package(url: "https://github.com/apple/swift-markdown.git", branch: "main"),
```
QuietMisdreavus, line 14: I don't see a `.package` method that has a `branch:` parameter in [the SwiftPM reference](https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#package-dependency). Is this something that you needed to do to get this to build on your system?

README.md
```diff
@@ -11,13 +11,13 @@ The markup tree provided by this package is comprised of immutable/persistent, t
...
 ```
 
 Add the dependency to any targets you've declared in your manifest:
 
 ```swift
-.target(name: "MyTarget", dependencies: ["Markdown"]),
+.product(name: "Markdown", package: "swift-markdown"),
```
QuietMisdreavus, line 20: While i think it's important to mention that the module name and package name are different, i feel like this change doesn't tell the whole story. How about combining both the original and your change?

```suggestion
.target(name: "MyTarget", dependencies: [
    .product(name: "Markdown", package: "swift-markdown"),
]),
```

QuietMisdreavus (changes requested):
Thanks for the PR! I had a couple comments about your changes.

README.md
mikekreuzer: It's there on the page you linked to :-)

Mostly I chose branch because that's what was there before, but there is a note on the pm reference that gives me pause:

/// Note that packages that use branch-based dependency requirements
/// can't be depended upon by packages that use version-based dependency
/// requirements; you should remove branch-based dependency requirements
/// before publishing a version of your package.

I don't know if the package manager really enforces that, it doesn't seem to, but we can't use version numbers here because the repo doesn't have semver numbered releases, just a lot of non semver tags.

QuietMisdreavus: I still don't see it as an overload of the `.package` method. Instead i see:

```swift
static func package(url: String, from version: Version) -> Package.Dependency
static func package(url: String, _ requirement: Package.Dependency.Requirement) -> Package.Dependency
static func package(url: String, _ range: Range<Version>) -> Package.Dependency
static func package(url: String, _ range: ClosedRange<Version>) -> Package.Dependency
static func package(path: String) -> Package.Dependency
```

The way to define a branch-based dependency is to use the `.branch` method on `Requirement` and the `.package(url:_:)` overload that takes a `Requirement`, like the existing readme states. Does the form you've added, `.package(url:branch:)` work for you? I'm really curious to see if there's an undocumented overload there.

ethan-kusters: I believe it's this one: https://developer.apple.com/documentation/packagedescription/package/dependency/package(url:branch:).

The `.branch` version (https://developer.apple.com/documentation/packagedescription/package/dependency/requirement-swift.enum/branch(_:)) is deprecated as of Swift 5.6 but still works.

README.md
mikekreuzer: Better, I agree

Kyle-Ye (approved):

Kyle-Ye:
Close in favor of #10

PR #79, opened 2022-09-25
Fix Directive argument name and value ranges on non-first line

### LINKED ISSUE #71: Directive argument name and value ranges are incorrect for single-line directives
## Summary
Directive argument name and value ranges are incorrect for single-line directives that are written anywhere besides the first line of the document.

## Details
I've reproduced this issue with the following test run on the current `main` branch:

```swift
func testDirectiveWithOneArgOnDifferentLines() throws {
    var source = """
        @Options(scope: page)
        """
    
    for line in 1...10 {
        let document = Document(parsing: source, options: .parseBlockDirectives)
        let directive = try XCTUnwrap(document.child(at: 0) as? BlockDirective)
        let arguments = directive.argumentText.parseNameValueArguments()
        let scopeArg = try XCTUnwrap(arguments["scope"])
        
        XCTAssertEqual("scope", scopeArg.name)
        XCTAssertEqual(
            SourceLocation(line: line, column: 10, source: nil)..<SourceLocation(line: line, column: 15, source: nil),
            scopeArg.nameRange,
            "Unexpected name range when authored at line '\(line)'"
        )
        
        XCTAssertEqual("page", scopeArg.value)
        XCTAssertEqual(
            SourceLocation(line: line, column: 17, source: nil)..<SourceLocation(line: line, column: 21, source: nil),
            scopeArg.valueRange,
            "Unexpected value range when authored at line '\(line)'"
        )
        
        source = "\n" + source
    }
}
```

which gives the following output:

```shell
XCTAssertEqual failed: ("Optional(Range(2:10..<2:15))") is not equal to ("Optional(Range(2:1..<2:6))") - Unexpected name range when authored at line '2'
XCTAssertEqual failed: ("Optional(Range(2:17..<2:21))") is not equal to ("Optional(Range(2:8..<2:12))") - Unexpected value range when authored at line '2'
XCTAssertEqual failed: ("Optional(Range(3:10..<3:15))") is not equal to ("Optional(Range(3:1..<3:6))") - Unexpected name range when authored at line '3'
XCTAssertEqual failed: ("Optional(Range(3:17..<3:21))") is not equal to ("Optional(Range(3:8..<3:12))") - Unexpected value range when authored at line '3'
XCTAssertEqual failed: ("Optional(Range(4:10..<4:15))") is not equal to ("Optional(Range(4:1..<4:6))") - Unexpected name range when authored at line '4'
XCTAssertEqual failed: ("Optional(Range(4:17..<4:21))") is not equal to ("Optional(Range(4:8..<4:12))") - Unexpected value range when authored at line '4'
XCTAssertEqual failed: ("Optional(Range(5:10..<5:15))") is not equal to ("Optional(Range(5:1..<5:6))") - Unexpected name range when authored at line '5'
XCTAssertEqual failed: ("Optional(Range(5:17..<5:21))") is not equal to ("Optional(Range(5:8..<5:12))") - Unexpected value range when authored at line '5'
XCTAssertEqual failed: ("Optional(Range(6:10..<6:15))") is not equal to ("Optional(Range(6:1..<6:6))") - Unexpected name range when authored at line '6'
XCTAssertEqual failed: ("Optional(Range(6:17..<6:21))") is not equal to ("Optional(Range(6:8..<6:12))") - Unexpected value range when authored at line '6'
XCTAssertEqual failed: ("Optional(Range(7:10..<7:15))") is not equal to ("Optional(Range(7:1..<7:6))") - Unexpected name range when authored at line '7'
XCTAssertEqual failed: ("Optional(Range(7:17..<7:21))") is not equal to ("Optional(Range(7:8..<7:12))") - Unexpected value range when authored at line '7'
XCTAssertEqual failed: ("Optional(Range(8:10..<8:15))") is not equal to ("Optional(Range(8:1..<8:6))") - Unexpected name range when authored at line '8'
XCTAssertEqual failed: ("Optional(Range(8:17..<8:21))") is not equal to ("Optional(Range(8:8..<8:12))") - Unexpected value range when authored at line '8'
XCTAssertEqual failed: ("Optional(Range(9:10..<9:15))") is not equal to ("Optional(Range(9:1..<9:6))") - Unexpected name range when authored at line '9'
XCTAssertEqual failed: ("Optional(Range(9:17..<9:21))") is not equal to ("Optional(Range(9:8..<9:12))") - Unexpected value range when authored at line '9'
XCTAssertEqual failed: ("Optional(Range(10:10..<10:15))") is not equal to ("Optional(Range(10:1..<10:6))") - Unexpected name range when authored at line '10'
XCTAssertEqual failed: ("Optional(Range(10:17..<10:21))") is not equal to ("Optional(Range(10:8..<10:12))") - Unexpected value range when authored at line '10'
```

showing that the name and value ranges are incorrect when the directive is authored beyond line 1.

## Notes

This only reproduces for single-line directives and so seems related to #49.

Kyle-Ye:
Bug/issue #, if applicable: 

Close #71 

## Summary

- Remove unnecessary optional check
- Fix and add test case for single line block parse

## Dependencies

None

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/Markdown/Base/DirectiveArgument.swift
+++ b/Sources/Markdown/Base/DirectiveArgument.swift
@@ -201,8 +201,8 @@ public struct DirectiveArgumentText: Equatable {
             var line = TrimmedLine(untrimmedText[...],
                                    source: range?.lowerBound.source,
                                    lineNumber: range?.lowerBound.line,
-                                   parseIndex: parseIndex,
-                                   startParseIndex: untrimmedText.index(parseIndex, offsetBy: 1 - (range?.lowerBound.column ?? 1)))
+                                   parseIndex: parseIndex
+            )
             line.lexWhitespace()
             while !line.isEmptyOrAllWhitespace {
                 let name: TrimmedLine.Lex?
--- a/Sources/Markdown/Base/DirectiveArgument.swift
+++ b/Sources/Markdown/Base/DirectiveArgument.swift
@@ -76,16 +76,6 @@ public struct DirectiveArgumentText: Equatable {
             self.range = range
         }
 
-        /// Returns a Boolean value indicating whether two line segments are equal.
-        /// - Parameter lhs: a line segment to compare
-        /// - Parameter rhs: another line segment to compare
-        /// - Returns: `true` if the two segments are equal.
-        public static func ==(lhs: LineSegment, rhs: LineSegment) -> Bool {
-            return lhs.untrimmedText == rhs.untrimmedText &&
-                lhs.parseIndex == rhs.parseIndex &&
-                lhs.range == rhs.range
-        }
-
         /// Parse a quoted literal.
         ///
         /// ```
--- a/Sources/Markdown/Base/DirectiveArgument.swift
+++ b/Sources/Markdown/Base/DirectiveArgument.swift
@@ -43,11 +43,14 @@ public struct DirectiveArgumentText: Equatable {
 
     /// A segment of a line of argument text.
     public struct LineSegment: Equatable {
-        /// The segment's untrimmed text from which arguments can be parsed.
+        /// The original untrimmed text of the line, from which arguments can be parsed.
         public var untrimmedText: String
 
-        /// The index in ``untrimmedText`` where the line started.
-        public var lineStartIndex: String.Index
+        @available(*, deprecated, renamed: "untrimmedText.startIndex")
+        public var lineStartIndex: String.Index {
+            get { untrimmedText.startIndex }
+            set { }
+        }
 
         /// The index from which parsing should start.
         public var parseIndex: String.Index
@@ -64,13 +67,11 @@ public struct DirectiveArgumentText: Equatable {
         /// Create an argument line segment.
         /// - Parameters:
         ///   - untrimmedText: the segment's untrimmed text from which arguments can be parsed.
-        ///   - lineStartIndex: the index in ``text`` where the line started.
-        ///   - parseIndex: index from which parsing should start.
+        ///   - parseIndex: The index from which parsing should start.
         ///   - range: The range from which a segment was extracted from a line
         ///     of source, or `nil` if the argument text was provided by other means.
-        init(untrimmedText: String, lineStartIndex: String.Index, parseIndex: String.Index? = nil, range: SourceRange? = nil) {
+        init(untrimmedText: String, parseIndex: String.Index? = nil, range: SourceRange? = nil) {
             self.untrimmedText = untrimmedText
-            self.lineStartIndex = lineStartIndex
             self.parseIndex = parseIndex ?? untrimmedText.startIndex
             self.range = range
         }
@@ -81,7 +82,6 @@ public struct DirectiveArgumentText: Equatable {
         /// - Returns: `true` if the two segments are equal.
         public static func ==(lhs: LineSegment, rhs: LineSegment) -> Bool {
             return lhs.untrimmedText == rhs.untrimmedText &&
-                lhs.lineStartIndex == rhs.lineStartIndex &&
                 lhs.parseIndex == rhs.parseIndex &&
                 lhs.range == rhs.range
         }
@@ -284,7 +284,7 @@ public struct DirectiveArgumentText: Equatable {
     /// from a string.
     public init<S: StringProtocol>(_ string: S) {
         let text = String(string)
-        self.segments = [LineSegment(untrimmedText: text, lineStartIndex: text.startIndex, range: nil)]
+        self.segments = [LineSegment(untrimmedText: text, range: nil)]
     }
 
     /// Create a body of argument text from a sequence of ``LineSegment`` elements.
--- a/Sources/Markdown/Base/DirectiveArgument.swift
+++ b/Sources/Markdown/Base/DirectiveArgument.swift
@@ -201,7 +201,8 @@ public struct DirectiveArgumentText: Equatable {
             var line = TrimmedLine(untrimmedText[...],
                                    source: range?.lowerBound.source,
                                    lineNumber: range?.lowerBound.line,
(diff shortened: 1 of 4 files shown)


Kyle-Ye:
IMG_3185

Some rough draft images that may help code review

Kyle-Ye:
@swift-ci please test

Sources/Markdown/Parser/BlockDirectiveParser.swift
```diff
@@ -265,12 +265,12 @@ struct TrimmedLine {
...
 
     /// - parameter untrimmedText: ``untrimmedText``
-    init(_ untrimmedText: Substring, source: URL?, lineNumber: Int?, parseIndex: Substring.Index? = nil) {
+    init(_ untrimmedText: Substring, source: URL?, lineNumber: Int?, parseIndex: Substring.Index? = nil, startParseIndex: Substring.Index? = nil) {
         self.untrimmedText = untrimmedText
         self.source = source
         self.parseIndex = parseIndex ?? untrimmedText.startIndex
         self.lineNumber = lineNumber
-        self.startParseIndex = self.parseIndex
+        self.startParseIndex = startParseIndex ?? self.parseIndex
```
QuietMisdreavus, line 273: This is useful, but i wonder if it would be better if the `location` property were able to scan backwards to the previous newline character (or the beginning of the string)? That way these calculations would be much more reliable.

Kyle-Ye: Should we only lazily scan once for the first time location is accessed? Or should we compute it every time? 

Overall, I'm -1 on back scan and delete startParseIndex property.

Since Markdown is a low level library I think performance is important in some way. And that's why we prefer to use index almost everywhere when parsing.

QuietMisdreavus: Part of me thinks that caching the `startParseIndex` in that situation wouldn't make much of a difference, since `TrimmedLine`s are generally created and dropped a lot in the parsing logic. And because of that, it's better for performance that we can use the index we already have on hand. I'll drop this comment.

Sources/Markdown/Base/DirectiveArgument.swift
```diff
@@ -201,7 +201,8 @@ public struct DirectiveArgumentText: Equatable {
             var line = TrimmedLine(untrimmedText[...],
                                    source: range?.lowerBound.source,
                                    lineNumber: range?.lowerBound.line,
-                                   parseIndex: parseIndex)
+                                   parseIndex: parseIndex,
+                                   startParseIndex: untrimmedText.index(parseIndex, offsetBy: 1 - (range?.lowerBound.column ?? 1)))
```
QuietMisdreavus, line 205: Looking at the code again, it seems like there's an existing `lineStartIndex` property on this `LineSegment` type that should always be populated. Did that not work for this purpose?

Kyle-Ye: `lineStartIndex` is init as `untrimmedText.startIndex` in BlockDirectParser.

eg.
```
(lldb) po untrimmedText
"hello\n @Options(scope: page"

(lldb) po untrimmedText[lineStartIndex]
"h"

(lldb) po untrimmedText[untrimmedText.index(parseIndex, offsetBy: 1 - (range?.lowerBound.column ?? 1))]
" "
```

Kyle-Ye:
The bug has been there for some time. Can anyone help continue to review this so that we can merge and solve it? Thanks

d-ronnqvist:
I looked at this today and as far as I can tell the issue is that `TrimmedLine` and `LineSegment` mean different things by `untrimmedText.` For `TrimmedLine` the untrimmed text is only expecting to hold one line of text:
```swift
/// The original untrimmed text of the line.
let untrimmedText: Substring
```
whereas `LineSegment` holds a potentially larger string and points to the index where the line starts
```swift
/// The segment's untrimmed text from which arguments can be parsed.
public var untrimmedText: String
/// The index in ``untrimmedText`` where the line started.
public var lineStartIndex: String.Index
```

Because of this difference, passing the full untrimmed line from the line segment to the trimmed line is probably not the right thing to do. 

I checked out this branch to continue investigating this issue and two things worth pointing out:
- The code that creates a `LineSegment` for each directive argument text doesn't compute `lineStartIndex`, it just passes the start index of the full text.
- The code that creates a `LineSegment` for each directive argument text _does_ slice of the "untrimmed text" after the end index of the argument text. For every line of directive arguments, this creates a copy of the entire document up to and including that line.
- The `TrimmedLine` only trims what comes before the `parseIndex`.

Since the start and end of the line segment's "untrimmedText" is sliced in two different places, it raises a question of which place should be responsible for slicing the line segment's text.

I think both of these should be done in one place, so I could imagine two different solutions here:
- Change `LineSegment` so that `untrimmedText` only stores one line of text (same as `TrimmedLine.untrimmedText`). This means that `lineStartIndex` is always `untrimmedText.startIndex`.
- Add API to `LineSegment` so that the line substring can be retrieved and use that so that `TrimmedLine` is initialized with only one line.

Looking at the other places that create a `LineSegment` I felt that the first alternative was the better solution, mainly because it makes both `untrimmedText` properties mean that same thing. 

Since I had already done most of this locally at this point I opened [a PR here][1] with the changes I made.

[1]: https://github.com/Kyle-Ye/swift-markdown/pull/1

Kyle-Ye:
Thanks for the follow-up. I'll read and review it later this week.

Kyle-Ye:
@swift-ci please test

Kyle-Ye:
Do we need a 3rd people to help review this? @d-ronnqvist Maybe you can invite one

daniel-grumberg (approved):

Kyle-Ye:
@d-ronnqvist Would you help give this a final review so that we can merge it?

d-ronnqvist (approved):

PR #82, opened 2022-10-08
Fix link missing range issue

Kyle-Ye:
Bug/issue #, if applicable: 

With https://github.com/apple/swift-docc/pull/376

Close https://github.com/apple/swift-docc/issues/271

## Summary

When updating destination of link, we preserve the old `parsedRange` instead of setting to nil.

The `parsedRange` is indeed out of date and we should not rely on it. But sometimes we need the original `parsedRange` to do some logic.

See discussion on https://github.com/apple/swift-docc/issues/271#issuecomment-1272279547

## Dependencies

None

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/Markdown/Inline Nodes/Inline Containers/Link.swift
+++ b/Sources/Markdown/Inline Nodes/Inline Containers/Link.swift
@@ -58,7 +58,7 @@ public extension Link {
             if let d = newValue, d.isEmpty {
                 _data = _data.replacingSelf(.link(destination: nil, parsedRange: nil, _data.raw.markup.copyChildren()))
             } else {
-                _data = _data.replacingSelf(.link(destination: newValue, parsedRange: nil, _data.raw.markup.copyChildren()))
+                _data = _data.replacingSelf(.link(destination: newValue, parsedRange: _data.range, _data.raw.markup.copyChildren()))
             }
         }
     }


Kyle-Ye:
@swift-ci please test

franklinsch:
@QuietMisdreavus are you happy with this change? It seems reasonable to me

Sources/Markdown/Inline Nodes/Inline Containers/Link.swift
```diff
@@ -58,7 +58,7 @@ public extension Link {
             if let d = newValue, d.isEmpty {
                 _data = _data.replacingSelf(.link(destination: nil, parsedRange: nil, _data.raw.markup.copyChildren()))
             } else {
-                _data = _data.replacingSelf(.link(destination: newValue, parsedRange: nil, _data.raw.markup.copyChildren()))
+                _data = _data.replacingSelf(.link(destination: newValue, parsedRange: _data.range, _data.raw.markup.copyChildren()))
```
franklinsch, line 61: Hm I worry a bit that Swift-Markdown is setting the `parsedRange` inconsistently across different markup elements. Can you explain why this change is needed? Is there another way clients could get the original range before changing the link's destination?

Kyle-Ye: > Can you explain why this change is needed?
Reason: See https://github.com/apple/swift-docc/issues/271#issuecomment-1272279547

> Is there another way clients could get the original range before changing the link's destination?

Other way IMO is to add a new property in Link so that we can differentiate []() and <doc:> in the consume site. Or we could add a new property `oldParsedRange` to store the oldRange every time we update the parsedRange.

franklinsch: I see. Adding a property to `Link` that tracks what kind of link it is (i.e., a `<>` 'auto-link' or a regular `[]()` link) seems totally reasonable to me. That way we don't need to fiddle with ranges.

PR #96, opened 2023-01-10
Adds support for "See Also" asides

ethan-kusters:
Bug/issue #, if applicable: rdar://58775869

“See Also” is a common kind of aside that folks parsing markdown may wish to consider.

For Swift-DocC specifically, there is a lot of pre-DocC Swift documentation content that uses this style of callout:

```
- SeeAlso:
```

In order to gracefully transition this content over to Swift-DocC style asides, we should add support for them. (There will need to be a follow-up PR in Swift-DocC to fully adopt this once this PR lands.)

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/Markdown/Interpretive Nodes/Aside.swift
+++ b/Sources/Markdown/Interpretive Nodes/Aside.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2023 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
@@ -92,6 +92,9 @@ public struct Aside {
         /// A "throws" aside.
         public static let `throws` = Kind(rawValue: "Throws")!
         
+        /// A "seeAlso" aside.
+        public static let seeAlso = Kind(rawValue: "SeeAlso")!
+        
         /// A collection of preconfigured aside kinds.
         public static var allCases: [Aside.Kind] {
             [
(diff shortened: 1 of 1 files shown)


Sources/Markdown/Interpretive Nodes/Aside.swift
```diff
@@ -118,9 +121,35 @@ public struct Aside {
                 todo,
                 version,
                 `throws`,
+                seeAlso,
```
ethan-kusters, line 124: Adding `seeAlso` to `allCases` adds test coverage for this new case in `AsideTests.swift`

Sources/Markdown/Interpretive Nodes/Aside.swift
```diff
@@ -118,9 +121,35 @@ public struct Aside {
...
+        ///
+        /// For multi-word asides this value may differ from the aside's ``rawValue``.
+        /// For example, the ``seeAlso`` aside's `rawValue` is `"SeeAlso"` but its
+        /// `displayName` is `"See Also"`.
+        /// Likewise, ``nonMutatingVariant``'s `rawValue` is
+        /// `"NonMutatingVariant"` and its `displayName` is `"Non-Mutating Variant"`.
+        ///
+        /// For simpler, single-word asides like ``bug``, the `displayName` and `rawValue` will
+        /// be the same.
+        public var displayName: String {
```
ethan-kusters, line 138: Swift-DocC currently implements this here: https://github.com/apple/swift-docc/blob/9c28bc4391e04bc6eaf4edf6226e1b94ee93805e/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift#L298.

I'm not sure it makes sense for Swift-Markdown to own the asides Swift-DocC knows how to parse but since that's the current state I think it makes sense for it to _also_ own the custom `displayName` values.

ethan-kusters:
@swift-ci please test

Sources/Markdown/Interpretive Nodes/Aside.swift
bitjammer: All absolutely fair statements. This was a difficult choice in the beginning but it just seemed too useful to other clients.

Sources/Markdown/Interpretive Nodes/Aside.swift
bitjammer: Thanks for the heads-up.

bitjammer (approved):
LG, thank you.

Sources/Markdown/Interpretive Nodes/Aside.swift
ethan-kusters: > This was a difficult choice in the beginning but it just seemed too useful to other clients.

That's a good point. Definitely see the value in Swift-Markdown being able to do this as well. It seems like we may have gone back and forth on it to some extent – I noticed there's also this simple list item extractor in Swift-DocC that's currently unused: https://github.com/apple/swift-docc/blob/9c28bc4391e04bc6eaf4edf6226e1b94ee93805e/Sources/SwiftDocC/Utility/MarkupExtensions/ListItemExtractor.swift#L18

QuietMisdreavus: It does seem odd to me that there's a split concern between Swift-Markdown and Swift-DocC, but it also seems like the list-item extractor in docc is mainly for extracting "Parameters" and "Returns" clauses, with it mostly deferring to Swift-Markdown for generic asides. (Though it does scan that list you linked, it seems like the only thing that happens is that list-items with those tags are dropped from the output entirely.)

QuietMisdreavus (approved):
Looks good, thanks!

PR #98, opened 2023-01-25
Pin version range of plugin dependency

bitjammer:
Start with the 1.1.0 version of the swift-docc-plugin dependency instead of the main branch, which may be unstable.


bitjammer:
@swift-ci Please test

QuietMisdreavus (approved):

PR #102, opened 2023-01-27
Create .spi.yml

finestructure:
As discussed with @franklinsch , this will ensure SPI documentation is updated without any particular workarounds in place in SPI.

We initially added custom support to host documentation on SPI without a `.spi.yml` file but this is becoming a bit tricky to maintain and has broken in a couple of instances.

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -1,0 +1,4 @@
+version: 1
+builder:
+  configs:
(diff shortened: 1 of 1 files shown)


franklinsch (approved):
Thank you!

franklinsch:
@swift-ci please test

PR #107, opened 2023-02-23
basic support for parsing Doxygen `\param` and `\returns` commands

QuietMisdreavus:
Bug/issue #, if applicable: rdar://69835334

## Summary

This PR introduces a new parser mode, adding support for a minimal set of Doxygen commands, to facilitate transitioning from alternate documentation systems into systems that use Swift-Markdown, such as Swift-DocC.

The parsing is gated behind a new parse option, currently named `parseMinimalDoxygen` (name open for bikeshedding). When both this and `parseBlockDirectives` are given, the block-directive parser will start looking for a specific set of Doxygen commands. This PR adds support for two basic commands: `\param` and `\returns` (though the latter will also recognize `\return` and `\result`, as those are synonyms in Doxygen itself).

To facilitate testing, this PR also extends the `markdown-tool dump-tree` command with an `--experimental-parse-doxygen-commands` flag which turns on the behavior.

I also fixed two other issues in the block directive parser in the course of this PR:

- If a document begins with a code fence, the contents of that code fence were incorrectly scanned for block directives.
- If a line run in a document or block directive included a line that didn't start on the beginning of the line (e.g. when a directive is written on one line), the source location was incorrectly given as the beginning of the line.

## Dependencies

This requires another PR to integrate with Swift-DocC, which is yet to be written.

## Testing

As the new functionality is gated behind a new ParseOption, testing should also ensure that existing functionality is not changed without the new option set.

Steps:
1. Write a Markdown file with Doxygen command content in it. See the tests added in this PR for examples.
2. `swift run markdown-tool dump-tree --parse-block-directives --experimental-parse-doxygen-commands test.md`
3. Inspect the result to ensure that it looks valid.
4. Repeat with `--source-locations` and ensure that the source locations for command descriptions look valid.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/Markdown/Block Nodes/Block Container Blocks/Doxygen Commands/DoxygenParam.swift
+++ b/Sources/Markdown/Block Nodes/Block Container Blocks/Doxygen Commands/DoxygenParam.swift
@@ -1,0 +1,78 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2023 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+
+/// A parsed Doxygen `\param` command.
+///
+/// The Doxygen support in Swift-Markdown parses `\param` commands of the form
+/// `\param name description`, where `description` extends until the next blank line or the next
+/// parsed command. For example, the following input will return two `DoxygenParam` instances:
+///
+/// ```markdown
+/// \param coordinate The coordinate used to center the transformation.
+/// \param matrix The transformation matrix that describes the transformation.
+/// For more information about transformation matrices, refer to the Transformation
+/// documentation.
+/// ```
+public struct DoxygenParam: BlockContainer {
+    public var _data: _MarkupData
+
+    init(_ raw: RawMarkup) throws {
+        guard case .doxygenParam = raw.data else {
+            throw RawMarkup.Error.concreteConversionError(from: raw, to: BlockDirective.self)
+        }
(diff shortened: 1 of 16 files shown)


QuietMisdreavus:
@swift-ci Please test

Sources/Markdown/Markdown.docc/Markdown/DoxygenCommands.md
```diff
@@ -0,0 +1,43 @@
+# Doxygen Commands
```
franklinsch, line 1: Nice docs!

Sources/Markdown/Block Nodes/Block Container Blocks/Doxygen Commands/DoxygenParam.swift
```diff
@@ -0,0 +1,78 @@
...
+/// `\param name description`, where `description` extends until the next blank line or the next
+/// parsed command. For example, the following input will return two `DoxygenParam` instances:
+///
+/// ```markdown
+/// \param coordinate The coordinate used to center the transformation.
+/// \param matrix The transformation matrix that describes the transformation.
+/// For more information about transformation matrices, refer to the Transformation
+/// documentation.
+/// ```
+public struct DoxygenParam: BlockContainer {
```
franklinsch, line 25: nit: Spelling this out as `DoxygenParameter` might be nicer. `param` is the lexical token that Doxygen uses, but we can refer to it as a "parameter" for clarity / consistency with Swift API naming guidelines.

franklinsch (approved):
This looks great. I had a thought that maybe we should be treating doxygen parsing as an implementation detail and exposing translated Markdown elements to clients, or to have a single `DoxygenCommand` model with an `arguments` property, but the approach you've taken is better from a client perspective because we interpret the Doxygen for them and we maintain fidelity when reformatting to a Markdown string.

QuietMisdreavus:
@swift-ci Please test

QuietMisdreavus:
@franklinsch Would you mind taking another look? Your comment reminded me that i hadn't added support in the markup formatter 😅

franklinsch:
Looks great, thanks!

PR #110, opened 2023-03-13
Add isAutolink property to Link

Kyle-Ye:
Bug/issue #, if applicable: 

With https://github.com/apple/swift-docc/pull/376

Close https://github.com/apple/swift-docc/issues/271

## Summary

- Add a new property isAutolink to `Link`
- Add a setter to `Link.isAutolink` so that after updating destination of link, we can preserve the old isAutolink value.

See discussion on https://github.com/apple/swift-docc/pull/376#discussion_r1133738397

## Dependencies

None

## Testing

https://github.com/apple/swift/pull/61506

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/Markdown/Inline Nodes/Inline Containers/Link.swift
+++ b/Sources/Markdown/Inline Nodes/Inline Containers/Link.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2023 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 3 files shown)


Kyle-Ye:
@swift-ci please test

Sources/Markdown/Inline Nodes/Inline Containers/Link.swift
```diff
@@ -63,6 +63,16 @@ public extension Link {
...
 
+    var isAutolink: Bool {
+        guard let destination,
+              childCount == 1,
+              let text = child(at: 0) as? Text,
+              destination == text.string else {
+            return false
+        }
+        return true
+    }
```
QuietMisdreavus, line 74: I evidently don't know enough about Swift to understand how this works. The test you wrote sets a value to this property, but i was under the assumption that computed properties like this could not be written to. How does this allow for a setter?

PR #116, opened 2023-03-29
consider queued newlines when wrapping elements in MarkupFormatter

QuietMisdreavus:
Bug/issue #, if applicable: rdar://106915293

## Summary

When an inline code span, image, link, or inline attribute is written directly after a soft or hard line break, the MarkupFormatter incorrectly adds an extra line break before the latter item. This leads to Markdown in these situations being incorrectly round-tripped, because now there is a paragraph break that wasn't there before. This PR makes the MarkupFormatter consider queued newlines in these checks.

This also fixes an error in the MarkupFormatter output for inline attributes, which were previously missing the caret that denoted them as inline attributes instead of links.

## Dependencies

None

## Testing

Use the following markdown file:

```markdown
This is some text.
`This is some code.`

This is some text.
[This is a link.](https://swift.org)

This is some text.
![This is an image.](image.png)

This is some text.
^[This is some attributed text.](rainbow: 'extreme')
```

Steps:
1. Save the previous Markdown as `test.md`.
2. `swift run markdown-tool format --check-structural-equivalence test.md`
3. Ensure that the rendered output contains only four paragraphs, and does not emit a "structural equivalence" warning.

## Checklist

- [x] Added tests
- [ n/a ] Updated documentation if necessary

### DIFF
--- a/Sources/Markdown/Walker/Walkers/MarkupFormatter.swift
+++ b/Sources/Markdown/Walker/Walkers/MarkupFormatter.swift
@@ -1138,7 +1138,7 @@ public struct MarkupFormatter: MarkupWalker {
     public mutating func visitInlineAttributes(_ attributes: InlineAttributes) {
         let savedState = state
         func printInlineAttributes() {
-            print("[", for: attributes)
+            print("^[", for: attributes)
             descendInto(attributes)
             print("](", for: attributes)
             print(attributes.attributes, for: attributes)
@@ -1152,7 +1152,7 @@ public struct MarkupFormatter: MarkupWalker {
         // gets into the realm of JSON formatting which might be out of scope of
         // this formatter. Therefore if exceeded, prefer to print it on the next
         // line to give as much opportunity to keep the attributes on one line.
-        if attributes.indexInParent > 0 && (isOverPreferredLineLimit || state.lineNumber > savedState.lineNumber) {
+        if attributes.indexInParent > 0 && (isOverPreferredLineLimit || state.effectiveLineNumber > savedState.effectiveLineNumber) {
             restoreState(to: savedState)
             queueNewline()
             printInlineAttributes()
--- a/Sources/Markdown/Walker/Walkers/MarkupFormatter.swift
+++ b/Sources/Markdown/Walker/Walkers/MarkupFormatter.swift
@@ -347,6 +347,11 @@ public struct MarkupFormatter: MarkupWalker {
 
         /// The current line number.
         var lineNumber = 0
+
+        /// The "effective" line number, taking into account any queued newlines which have not been printed.
+        var effectiveLineNumber: Int {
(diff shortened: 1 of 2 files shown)


QuietMisdreavus:
@swift-ci Please test

Sources/Markdown/Walker/Walkers/MarkupFormatter.swift
```diff
@@ -347,6 +347,11 @@ public struct MarkupFormatter: MarkupWalker {
 
         /// The current line number.
         var lineNumber = 0
+
+        /// The "effective" line number, taking into account any queued newlines which have not been printed.
```
ethan-kusters, line 351: Just for my understanding – what value does the `lineNumber` property provide? It seems confusing to have both of these now. I wouldn't be sure which to use in a given scenario. Maybe we can expand the docs of `lineNumber` to explain where its useful? It seems like its doc comment, "The current line number", is maybe not accurate?

(I'm wondering if we should rename it to something like `lastLineNumberWithContent` but that's a little verbose and I imagine would cause a fair bit of churn in the repo...)

QuietMisdreavus: Looking at the implementation, the `lineNumber` property is meant to represent the last line of *printed content*, and line breaks are queued up to be printed the next time non-blank content is printed. (I'm not sure the precise reason for this design decision, but it may have something to do with ensuring that line prefixes and leading indentation are correctly processed? This code is relatively old so i'd have to dig to find the original author.)

I'll update the doc comment to reflect the way `lineNumber` and `effectiveLineNumber` are used.

QuietMisdreavus:
@swift-ci Please test

Sources/Markdown/Walker/Walkers/MarkupFormatter.swift
```diff
@@ -345,8 +345,18 @@ public struct MarkupFormatter: MarkupWalker {
         /// The length of the last line.
         var lastLineLength = 0
 
-        /// The current line number.
+        /// The line number of the most recently printed content. This is updated in
+        /// `addressPendingNewlines(for:)` when line breaks are printed.
```
ethan-kusters, line 349: Nit: Maybe this should be in the discussion section?

```suggestion
        /// The line number of the most recently printed content. 
        ///
        /// This is updated in `addressPendingNewlines(for:)` when line breaks are printed.
```

ethan-kusters (approved):
Looks great!

Sources/Markdown/Walker/Walkers/MarkupFormatter.swift
QuietMisdreavus: I like this idea! I went ahead and applied your suggestion.

QuietMisdreavus:
@swift-ci Please test

PR #135, opened 2023-07-06
Add primary associated type for MarkupVisitor

Kyle-Ye:
Bug/issue #, if applicable: 

## Summary

Add primary associated type for `MarkupVisitor`

Update: Bump the minimal supported Swift version to Swift 5.7

## Alternatives

~The current implementation of PAT support for `MarkupVisitor` has a lot of duplicated code since PAT was introduced by Swift 5.7.~

~We could drop such duplication if we can update this Package's min supported Swift version to Swift 5.7.~

See #78

## Testing

See `MarkupVisitorTests.swift`

## Checklist

- [x] Added tests

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.5
+// swift-tools-version:5.7
 /*
  This source file is part of the Swift.org open source project
 
--- a/Sources/Markdown/Visitor/MarkupVisitor.swift
+++ b/Sources/Markdown/Visitor/MarkupVisitor.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2023 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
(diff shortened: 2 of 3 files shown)


QuietMisdreavus:
I think there's a broader discussion to be had about upgrading the minimum Swift version for Swift-DocC and its dependencies. This could probably be a Documentation Workgroup topic, IMO.

Kyle-Ye:
> I think there's a broader discussion to be had about upgrading the minimum Swift version for Swift-DocC and its dependencies. This could probably be a Documentation Workgroup topic, IMO.

If we can upgrade min-supported version to Swift 5.7. I think this PR is fairly easy to be merged.

If we decide to maintain a Swift version lower than Swift 5.7, it looks like we need to maintain such huge duplication of code to get PAT support.

Do you think this PR is generally fine even if we still need to target a swift version lower than 5.7?

Kyle-Ye:
Follow the swift repo's style, I just added 2 labels for this repo so that we can better tag this for the future PR.

image

Kyle-Ye:
> I think there's a broader discussion to be had about upgrading the minimum Swift version for Swift-DocC and its dependencies. This could probably be a Documentation Workgroup topic, IMO.

The Swift 5.7 bump was discussed and got approved at [last week's meeting](https://forums.swift.org/t/documentation-workgroup-meeting-july-17th-2023/66170/3
). 

I think this PR is now ready for review. @QuietMisdreavus

Kyle-Ye:
For PAT support, it unlocks more usage for the downstream user. Basically the same as Combine add PAT for Publisher protocol Publisher<Output, Failure>
Before the PR, we can only use `some MarkupVisitor` and can't do anything for the result of visit method since it is unbound.
```swift
var vistor: some MarkupVisitor = EmptyWalker()
let result = vistor.visit(Text(""))
// Can't do anything on result
```
After the PR, we can unlock something very interesting for the downstream user. See the latest test case. 84a204890458a4f9b35b990e8e6eaae6226e043b
(The original test case use `MarkupWalker` whose Result type is Void and may not fully express the idea. I've update the test case to show the PAT's usage)

You may also have a check here https://www.swiftbysundell.com/articles/opaque-return-types-primary-associated-types/ which has a great explanation for PAT.

Kyle-Ye:
@swift-ci please test

Kyle-Ye:
@swift-ci please test

d-ronnqvist (approved):

Kyle-Ye:
@swift-ci please test

PR #141, opened 2023-08-14
build: add a CMake based build

compnerd:
This is meant to support building swift-markdown for re-use in the toolchain build.  It does not follow best practices (opting for a single file approach rather than a CMakeLists.txt per directory) and currently needs some refinement for integrating an existing CMark build, but provides the skeleton for further experimentation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,61 @@
+#[[
+This source file is part of the Swift System open source project
+
+Copyright (c) 2024 Apple Inc. and the Swift System project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
+#]]
+
+cmake_minimum_required(VERSION 3.19.0)
+
+if(POLICY CMP0077)
+  cmake_policy(SET CMP0077 NEW)
+endif()
+
+if(POLICY CMP0091)
+  cmake_policy(SET CMP0091 NEW)
+endif()
+
+project(SwiftMarkdown
+  LANGUAGES C Swift)
+
+list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
+
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+
+set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
+set(CMAKE_Swift_COMPILE_OPTIONS_MSVC_RUNTIME_LIBRARY MultiThreadedDLL)
+
+include(FetchContent)
+include(SwiftSupport)
+
+set(_SM_VENDOR_DEPENDENCIES)
+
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
...
+find_package(ArgumentParser CONFIG REQUIRED)
+include(cmark-gfm)
+
+add_library(CAtomic STATIC
+  Sources/CAtomic/CAtomic.c)
+target_include_directories(CAtomic PUBLIC
+  Sources/CAtomic/include)
+
+add_library(Markdown
+  Sources/Markdown/Base/ChildIndexPath.swift
```
Kyle-Ye, line 31: Can we use the folder inclusion here? This kind of file enumeration scheme does not look reliable and may be broken in the future.

compnerd: No, globs are not really reliable and make using it far more difficult as they do not pickup changes as they are performed only once.  The CMake recommendation is explicit file lists.

Kyle-Ye: Got it. Thanks for the clarification.

compnerd:
CC: @QuietMisdreavus - the recent work in swift-cmark is in support of this change

compnerd:
@swift-ci please test

CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+
+include(FetchContent)
+include(SwiftSupport)
+
+set(_SM_VENDOR_DEPENDENCIES)
+
+set(BUILD_EXAMPLES NO)
+set(BUILD_TESTING NO)
+
+find_package(ArgumentParser CONFIG)
```
bnbarham, line 40: Where are the ArgumentParser and cmark-gfm targets aliased?

CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
+#[[
+This source file is part of the Swift System open source project
+
+Copyright (c) 2020 Apple Inc. and the Swift System project authors
```
bnbarham, line 4: 2020 -> 2024 😅?

CMakeLists.txt
Kyle-Ye: They are here.

- https://github.com/apple/swift-cmark/pull/69
- https://github.com/apple/swift-argument-parser/blob/main/CMakeLists.txt

Could you please add the PR's dependency to the body here? @compnerd 

(See the PR template here: https://github.com/apple/swift-markdown/blob/main/.github/pull_request_template.md)

bnbarham: To be clear, I know where the repositories are - this comment is about `find_package` namespacing its target names, where as `FetchContent` will not. See eg. https://github.com/apple/swift-format/pull/677/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR76 for what I mean.

CMakeLists.txt
compnerd: What year are we in? Is it not 1897? 😅

CMakeLists.txt
compnerd: Ugh, I think I missed that part of the change, I'll fix that up.

compnerd: Ah, swift-argument-parser and cmark both do not namespace their export targets, so there is no need for the aliasing.

compnerd: @Kyle-Ye - there are no dependencies for this change, it is fully standalone, it just adds the CMake build rules.

compnerd:
@swift-ci please test

bnbarham (approved):

PR #146, opened 2023-09-10
add Hashable, Sendable

zunda-pixel:
Bug/issue #, if applicable: 

## Summary

add Hashable, Sendable to Aside.Kind

## Dependencies

None

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/Markdown/Interpretive Nodes/Aside.swift
+++ b/Sources/Markdown/Interpretive Nodes/Aside.swift
@@ -22,7 +22,7 @@ import Foundation
 /// ```
 public struct Aside {
     /// Describes the different kinds of aside.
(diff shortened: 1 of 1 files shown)


zunda-pixel:
can you review @QuietMisdreavus ?

Kyle-Ye:
Could you add some motivation points for adding such conformance?

I'm curious what it unlocks at this point. And it may cause problem if the downstream user of Markdown declared the conformance on their side.(Although they shouldn't do this, because Hashable and Kind both are not the types they control.)

zunda-pixel:
I want to use in SwiftUI.
ForEach requires that Element conforms Hashable

```swift
ForEach(Aside.Kind.allCases, id: \.self) { kind in
  Text(kind.rawValue)
}
```

PR #151, opened 2023-10-05
Fix multi line symbol link source range issue

### LINKED ISSUE #142: Unexpected symbol link source range
## Summary 

When parsing a symbol link with multiline content, the source range is unexpected, new lines are replaced by whitespaces. 

For instance, parsing the following symbol link:

```
Test of a ``multi
line symbolink``
```

gives a symbol link with a range of `1:11-1:35`. Which is the equivalent of:

```
Test of a ``multi line symbolink``
```

 However, I would rather expect either:

- A range of `1:11-2:17`
- An invalid symbol link

## Steps to reproduce

```swift
let source = """
Test of a ``multi
line symbolink``
"""
let document = Document(parsing: source, options: .parseSymbolLinks)
print(document.debugDescription(options: .printEverything))
```

```
Document @1:1-2:17 Root #1 #0
└─ Paragraph @1:1-2:17 #1
   ├─ Text @1:1-1:11 #2 "Test of a "
   └─ SymbolLink @1:11-1:35 #3 destination: multi line symbolink
```
## Notes

This is a suspected cause of a crash in DocC diagnostic formatter. https://github.com/apple/swift-docc/issues/678#issuecomment-1684914893

Kyle-Ye:
Bug/issue #, if applicable: 

Close #142

## Summary

Fix the unexpected source range issue for multi line symbol link

For detail see https://github.com/apple/swift-markdown/issues/142#issuecomment-1748504438

2 question to be discussed:
1. Should we add CMARK_OPT_SOURCEPOS by default the same as CMARK_OPT_TABLE_SPANS?
2. Should we add a new ParseOptions case so that client of Markdown can choose to opt-in or opt-out this option?

## Dependencies

None

## Testing

Add testMultilineSymbolLink test case

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/Markdown/Parser/CommonMarkConverter.swift
+++ b/Sources/Markdown/Parser/CommonMarkConverter.swift
@@ -611,7 +611,7 @@ struct MarkupParser {
     static func parseString(_ string: String, source: URL?, options: ParseOptions) -> Document {
         cmark_gfm_core_extensions_ensure_registered()
 
-        var cmarkOptions = CMARK_OPT_TABLE_SPANS
+        var cmarkOptions = CMARK_OPT_TABLE_SPANS | CMARK_OPT_SOURCEPOS
         if !options.contains(.disableSmartOpts) {
             cmarkOptions |= CMARK_OPT_SMART
         }
--- a/Sources/Markdown/Parser/CommonMarkConverter.swift
+++ b/Sources/Markdown/Parser/CommonMarkConverter.swift
@@ -611,9 +611,12 @@ struct MarkupParser {
     static func parseString(_ string: String, source: URL?, options: ParseOptions) -> Document {
         cmark_gfm_core_extensions_ensure_registered()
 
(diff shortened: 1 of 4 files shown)


Kyle-Ye:
@swift-ci please test

d-ronnqvist (approved):

d-ronnqvist:
> 1. Should we add CMARK_OPT_SOURCEPOS by default the same as CMARK_OPT_TABLE_SPANS?
> 2. Should we add a new ParseOptions case so that client of Markdown can choose to opt-in or opt-out this option?

From what I understand the performance impact of `CMARK_OPT_SOURCEPOS` should we could enable it by default.

Regardless if we enable it by default or not, adding a ParseOption so that callers can opt in or out would be good.

Kyle-Ye:
Added the ParseOption.

Another thing that confuses me is that CMARK_OPT_SOURCEPOS seems to have other side effects in addition to solving this problem.

image

Kyle-Ye:
@swift-ci please test

PR #164, opened 2024-02-13
Support Windows ARM64 builds

hjyamauchi:
## Summary

Handle the "ARM64" system processor type.

## Dependencies

none

## Testing

Build the windows arm64 toolchain with build.ps1.

Steps:
1. _Provide setup instructions._
2. _Explain in detail how the functionality can be tested._

## Checklist

- [n] Added tests
- [n] Ran the `./bin/test` script and it succeeded
- [n] Updated documentation if necessary


hjyamauchi:
@compnerd

compnerd (approved):

Kyle-Ye:
I see some similar PR here https://github.com/apple/swift-corelibs-libdispatch/pull/803

Should we keep them in sync? cc @compnerd

compnerd:
I think that the patchset ended up with a second clause, which seems less likely to change existing behaviour. I think that trying to keep these files in sync is a great thing, but need to figure out how to do that still.

PR #167, opened 2024-02-21
build: silence warning about CMakeLists.txt from SPM

compnerd:
Ignore the CMakeLists.txt which is used for the CMake build to support swift-format.

Fixes: #166

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -28,6 +28,8 @@ let package = Package(
                 "CAtomic",
                 .product(name: "cmark-gfm", package: cmarkPackageName),
                 .product(name: "cmark-gfm-extensions", package: cmarkPackageName),
+            ], exclude: [
+                "CMakeLists.txt"
             ]),
         .testTarget(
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

d-ronnqvist (approved):

Package.swift
```diff
@@ -28,6 +28,8 @@ let package = Package(
                 "CAtomic",
                 .product(name: "cmark-gfm", package: cmarkPackageName),
                 .product(name: "cmark-gfm-extensions", package: cmarkPackageName),
+            ], exclude: [
```
Kyle-Ye, line 31: Nit: add a new line to align the style

```suggestion
            ], 
            exclude: [
```

Kyle-Ye (commented):
Should we also need to add the exclude parameter for "CAtomic" taget?

compnerd:
@swift-ci please test

PR #213, opened 2025-01-30
build: adjust the build on Windows to avoid some warnings

compnerd:
We explicitly build cmark-gfm as static. This adjusts the usage to ensure that the module knows that the cmark that it is building against is static. This fixes a number of linker warnings on Windows.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -28,9 +28,13 @@ let package = Package(
                 "CAtomic",
                 .product(name: "cmark-gfm", package: cmarkPackageName),
                 .product(name: "cmark-gfm-extensions", package: cmarkPackageName),
-            ], 
+            ],
             exclude: [
                 "CMakeLists.txt"
+            ],
+            swiftSettings: [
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

QuietMisdreavus (approved):
🚢

tonyarnold:
This change now means that we can't use the project as a dependency within Xcode: 

```
the target 'Markdown' in product 'Markdown' contains unsafe build flags
```

dlskawns96:
I’m also running into the same issue when trying to depend on swift-markdown from an iOS project in Xcode.  
As mentioned, SwiftPM rejects the package with:

> the target 'Markdown' in product 'Markdown' contains unsafe build flags

To address this, I opened #239 describing the problem, and submitted PR #240 that guards the `unsafeFlags` behind a Windows-only conditional, so that non-Windows consumers are unaffected.

It would be great to get feedback on whether that approach makes sense to merge.

/cc @compnerd @QuietMisdreavus

PR #221, opened 2025-04-17
Make types Sendable that are useful for AST inspection and library usage

DivineDominion:
Bug/issue #, if applicable: #170 

## Summary

In Swift 6 language mode, useful descriptions of the AST like `SourceLocation` and `ChildIndexPath`, but also the initial `ParseOptions`, aren't sendable.

While markup cannot be sendable (because of RawMarkup), a lot of types describing parts of the document can. This allows package users to work with e.g. the `SourceRange` of a markup element directly, for example in editors.

## Checklist

- [ ] Added tests -- nothing new to test


QuietMisdreavus (approved):
Nice catch, thanks!

QuietMisdreavus:
@swift-ci Please test

PR #129, opened 2023-05-31
Footnotes

NioDeBacker:
## Summary

Added footnote parsing functionality. Consists of two parts: Footnote Reference and Footnote Definition

## Dependencies

It is dependend on this version of swift-cmark: https://github.com/apple/swift-cmark/pull/57

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran the `./bin/test` script and it succeeded
- [n/a ] Updated documentation if necessary

PR #228, opened 2025-05-08
Add footnotes.

jbeder:
This relies on the footnote extension in cmark, which does not expose footnote type names, but does expose enough information via the type enum.

Bug/issue #, if applicable: 

## Summary

Adds new footnote types FootnoteDefinition and FootnoteReference to the markdown tree.

## Dependencies

None. It would be *nice* to use https://github.com/swiftlang/swift-cmark/pull/57, but it's not necessary; there's a small hack that can reliably detect when cmake produces footnotes.

This is heavily based on https://github.com/swiftlang/swift-markdown/pull/129, which also adds footnote support; however, that PR *does* rely on the upstream swift-cmark PR, whereas this one does not.

## Testing

New test introduced to verify the parsing of footnotes, which pass with `./bin/test`.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/Markdown/Base/Markup.swift
+++ b/Sources/Markdown/Base/Markup.swift
@@ -71,6 +71,10 @@ func makeMarkup(_ data: _MarkupData) -> Markup {
         return SymbolLink(data)
     case .inlineAttributes:
         return InlineAttributes(data)
+    case .footnoteReference:
+        return FootnoteReference(data)
+    case .footnoteDefinition:
+        return FootnoteDefinition(data)
     case .doxygenDiscussion:
         return DoxygenDiscussion(data)
     case .doxygenNote:
--- a/Sources/Markdown/Base/RawMarkup.swift
+++ b/Sources/Markdown/Base/RawMarkup.swift
@@ -52,6 +52,9 @@ enum RawMarkupData: Equatable {
     case tableRow
     case tableCell(colspan: UInt, rowspan: UInt)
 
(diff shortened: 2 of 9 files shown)


jbeder:
Hi, I just wanted to check in about this PR? I rebased on top of the latest, and fixed a couple of typos. It would be nice to get this merged to avoid having to maintain a separate fork. Happy to make any changes you want.

rsalesas:
Independent data point, in case it's useful for review: I hit the same gap and wrote footnote support before I found this PR, and landed on the same design — `FootnoteDefinition` as a block container and `FootnoteReference` as an inline, both carrying the author's identifier, and the same workaround for `cmark_node_get_type_string` not reporting footnote types (asking `cmark_node_get_type` instead, since the enum constants are public even where the strings aren't).

Two details from doing it that might be worth checking against this PR:

- cmark's iterator doesn't classify a footnote reference as a leaf — it emits both ENTER and EXIT for one — so it has to be converted as a container that happens to have no children. Treating it as a leaf trips the `cmark iterators should never return EXIT events for leaf nodes` precondition.
- A reference's own literal isn't its label. `process_footnotes` overwrites it with the assigned number, so the identifier has to come from `cmark_node_parent_footnote_def`.

Mine is at https://github.com/rsalesas/swift-markdown if any of it is useful. I'm not putting it forward as an alternative — this PR is older and I'd rather see it land than add a third to the queue. Happy to help with review or testing.

PR #259, opened 2026-03-11
resolves #93 - allows backticks in code voice

### LINKED ISSUE #93: Support for single backtick (`) in code voice
This issue appears is input like the following:

```
> *identifier* → **`` ` ``** *identifier-head* *identifier-characters*_?_ **`` ` ``**
```

The [CommonMark](https://spec.commonmark.org/0.30/#code-spans) definition of a code voice permits one or more backticks, which supports exactly this sort of scenario where a backtick needs to appear in code voice.  Additionally, [escaping via backslash](https://spec.commonmark.org/0.30/#backslash-escapes) isn't allowed in code voice.

The issue here is a collision with DocC's extended markup using double-backticks surround links to API symbols.  Currently, building "The Swift Programming Language" with top-of-tree `docc`, the output is correct but we get warnings like the following:

> Topic reference '`' couldn't be resolved. No local documentation matches this reference.

Trying 3 or 4 backticks to start/end the code voice produces the same result.

heckj:
resolves #93 - which allows backticks to appear in "code voice", used in content within https://github.com/swiftlang/swift-book (see https://github.com/swiftlang/swift-book/issues/71 for some detail in that repo)

This fix leans a concept that symbols when done with backticks, will always use exactly 2 backticks - and adds a smidge of a double check to make sure that the symbol content itself doesn't contain a backtick, which *I think* is a safe bet.

### DIFF
--- a/Sources/Markdown/Parser/CommonMarkConverter.swift
+++ b/Sources/Markdown/Parser/CommonMarkConverter.swift
@@ -403,7 +403,8 @@ struct MarkupParser {
         let parsedRange = state.range(state.node)
         let literalContent = getLiteralContent(node: state.node)
         if state.options.contains(.parseSymbolLinks),
-           cmark_node_get_backtick_count(state.node) > 1 {
+           cmark_node_get_backtick_count(state.node) == 2,
+           !literalContent.contains("`") {
             return MarkupConversion(state: state.next(), result: .symbolLink(parsedRange: parsedRange, destination: literalContent))
         } else {
             return MarkupConversion(state: state.next(), result: .inlineCode(parsedRange: parsedRange, code: literalContent))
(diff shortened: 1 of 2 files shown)


heckj:
@swift-ci please test

Sources/Markdown/Parser/CommonMarkConverter.swift
```diff
@@ -403,7 +403,8 @@ struct MarkupParser {
         let parsedRange = state.range(state.node)
         let literalContent = getLiteralContent(node: state.node)
         if state.options.contains(.parseSymbolLinks),
-           cmark_node_get_backtick_count(state.node) > 1 {
+           cmark_node_get_backtick_count(state.node) == 2,
```
QuietMisdreavus: Should this condition still be a `> 1`? We currently allow symbol links with more than two backticks so i'd rather keep that behavior if possible.

heckj: I thought it was safe to say symbols are only ever double-back-ticks, but the code works with just checking if the interior content includes a ` itself, so that does pass. The flip side of that is that any docc symbols can't ever include a backtick, and I wasn't sure if that was a super safe bet or not.

I'll push an update that drops back to `> 1`

heckj:
@swift-ci please test

QuietMisdreavus (approved):
This looks great, thanks!

heckj:
@swift-ci please test

heckj:
@swift-ci please test

PR #261, opened 2026-04-05
Add compatibility for Embedded Swift

MaxDesiatov:
## Summary

When building `swift-markdown` with Embedded Swift for Wasm, multiple issues occur:
1. Dependency on Foundation doesn't allow building for Embedded Swift, but this dependency is not required for most Embedded Swift use cases
2. Recursive generics with `ManagedBuffer` case compiler hangs, which should emit error diagnostics instead, fixed in https://github.com/swiftlang/swift/pull/88296

## Dependencies

https://github.com/swiftlang/swift/pull/88296 is not a strict dependency, since it will only cause the build of `main` to fail instead of hang. We're working around the use of recursive `ManagedBuffer` when building for Embedded Swift altogether in this PR.

## Testing

Existing non-embedded builds test suite passes. Since Swift Testing doesn't support Embedded Swift yet, this configuration is verified manually.

## Checklist

- [ ] Added tests

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,6 +16,7 @@ let cmarkPackageName = ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DE
 
 let package = Package(
     name: "swift-markdown",
+    platforms: [.macOS(.v10_15)],
     products: [
         .library(
             name: "Markdown",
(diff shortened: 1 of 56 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

PR #272, opened 2026-06-02
Fix a crash when the aside tag includes an en-dash and has no text afyer the tag-separator

d-ronnqvist:
Bug/issue #, if applicable: rdar://178409822

## Summary

This fixes a crash when an aside tag contains `--` or `---` _and_ has no further _text_ after then tag separator (`:`).

For example, this markup crashes
```md
> Before -- after:
> - Some list item
```

but this modified example doesn't because the tag includes a hyphen instead of a en-dash:
```md
> Before - after:
> - Some list item
```

and this other modified example doesn't because there is more text after the tag separator (`:`)
```md
> Before - after: some text
> - Some list item
```

## Dependencies

None.

## Testing

Parse markup like 
```md
> Before -- after:
> - Some list item
```
or 
```md
> Before --- after:
> - Some list item
```
as an aside. It shouldn't crash with a `Range requires lowerBound <= upperBound` failure.

## Checklist

- [x] Added tests
- ~[ ] Updated documentation if necessary~

### DIFF
--- a/Tests/MarkdownTests/Interpretive Nodes/AsideTests.swift
+++ b/Tests/MarkdownTests/Interpretive Nodes/AsideTests.swift
@@ -216,6 +216,38 @@ class AsideTests: XCTestCase {
 
         XCTAssertEqual(expectedRootDump, aside.content[0].root.debugDescription(options: .printSourceLocations))
     }
+    
+    func testDoesNotCrashParsingTagWithEnDashAndListItemContent() throws {
+        let enDash = "–"
+        let emDash = "—"
+        
+        for (dashMarkup, expectedColumn, formattedDash) in [
+            ("-",   18, "-"),
+            ("--",  19, enDash),
+            ("---", 20, emDash),
+            
+            (enDash, 20, enDash), // The unicode en-dash character is 3 UTF-8 bytes
+            (emDash, 20, emDash), // The unicode em-dash character is 3 UTF-8 bytes
+        ]{
+            try assertAside(
+                source: """
+                > Before \(dashMarkup) after:
+                > - All other content is inside unordered list
+                """,
+                conversionStrategy: .tagNotRequired,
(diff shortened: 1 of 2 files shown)


d-ronnqvist:
@swift-ci please test

Tests/MarkdownTests/Interpretive Nodes/AsideTests.swift
```diff
@@ -216,6 +216,38 @@ class AsideTests: XCTestCase {
 
         XCTAssertEqual(expectedRootDump, aside.content[0].root.debugDescription(options: .printSourceLocations))
     }
+    
+    func testDoesNotCrashParsingTagWithEnDashAndListItemContent() throws {
+        let enDash = "–"
+        let emDash = "—"
```
snprajwal: Nit: with a monospaced font, both these characters look the same. I verified that they're the right characters by copying them into a non-monospace text box and visually checking the width, but maybe it's better to use unicode escape sequences to make this more obvious here?


```suggestion
        let enDash = "\u{2013}" // U+2013 en-dash
        let emDash = "\u{2014}" // U+2014 em-dash
```

snprajwal (approved):
Nice one, thanks!

Tests/MarkdownTests/Interpretive Nodes/AsideTests.swift
d-ronnqvist: Thanks. Yes, I like how the unicode literals makes it clear what these characters are. 

I updated it to use unicode literals without the comments in https://github.com/swiftlang/swift-markdown/pull/272/commits/7c9f27821e27f627cd7c3782656aa5554fac147a
I didn't feel that the code comments added much because the variable names are already mentioned the character that each variable represents and their respective unicode literal already shows the numeric value for that character.

d-ronnqvist:
@swift-ci please test

PR #273, opened 2026-06-02
[6.4] Fix a crash when the aside tag includes an en-dash and has no text after the tag-separator

d-ronnqvist:
- **Explanation**: Fix a crash when the aside tag includes an en-dash and has no text after the tag-separator
- **Scope**: Resolves an edge case crash when an aside tag includes and en-dash, written as two hyphens (`--`), and the tag has no _text_ after the tag separator (`:`)—but instead has an image, unordered list, etc.
- **Issues**: rdar://178409822
- **Original PRs**: #272
- **Risk**: Low. Isolated bounds in the computation of the aside's tag range and remaining text range.
- **Testing**: See testing notes on #272
- **Reviewers**: @snprajwal


franklinsch (approved):

d-ronnqvist:
@swift-ci please test

PR #274, opened 2026-06-05
Escape HTML special characters in HTMLFormatter

### LINKED ISSUE #264: Bug: Html formatter parses html escaped code
Given this markdown file:

`&lt;key&gt;`

The html output should display as:

&lt;key&gt;

However, the current HTML formatter uses `text.string`:

https://github.com/swiftlang/swift-markdown/blob/55d66d9a9e8d4fd3f48d111b0d437e82fe451903/Sources/Markdown/Walker/Walkers/HTMLFormatter.swift#L274-L276

This will interpret the input as html data and thus produces an actual html tag (which will be hidden from view)

somiljain2006:
Bug/issue #, if applicable: Fixes #264

## Summary

The HTML formatter previously emitted text, inline code, and code block contents without escaping HTML special characters. As a result, markdown containing escaped entities such as:
```markdown
&lt;key&gt;
```
could be rendered as HTML that browsers interpret as an actual tag rather than displaying the text <key>. This change escapes &, <, and > when rendering text, inline code, and code block contents in HTMLFormatter. Raw HTML nodes (HTMLBlock and InlineHTML) continue to be emitted unchanged.

## Dependencies

N/A

## Testing

Regression tests have been added to verify correct rendering for:

- Regular text containing escaped HTML entities
- Inline code containing angle brackets
- Code blocks containing angle brackets

Steps:
1. Build and run the test suite:
2. Verify the new regression tests pass.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/Markdown/Walker/Walkers/HTMLFormatter.swift
+++ b/Sources/Markdown/Walker/Walkers/HTMLFormatter.swift
@@ -264,9 +264,6 @@ public struct HTMLFormatter: MarkupWalker {
         if let destination = link.destination {
             result += " href=\"\(destination)\""
         }
-        if let title = link.title, !title.isEmpty {
-            result += " title=\"\(title)\""
-        }
         result += ">"
 
         descendInto(link)
@@ -331,7 +328,5 @@ private extension String {
             .replacingOccurrences(of: "&", with: "&amp;")
             .replacingOccurrences(of: "<", with: "&lt;")
             .replacingOccurrences(of: ">", with: "&gt;")
-            .replacingOccurrences(of: "\"", with: "&quot;")
-            .replacingOccurrences(of: "'", with: "&#39;")
     }
 }
--- a/Sources/Markdown/Walker/Walkers/HTMLFormatter.swift
+++ b/Sources/Markdown/Walker/Walkers/HTMLFormatter.swift
@@ -86,7 +86,7 @@ public struct HTMLFormatter: MarkupWalker {
         } else {
             languageAttr = ""
         }
-        result += "<pre><code\(languageAttr)>\(codeBlock.code)</code></pre>\n"
+        result += "<pre><code\(languageAttr)>\(codeBlock.code.htmlEscaped())</code></pre>\n"
     }
 
     public mutating func visitHeading(_ heading: Heading) -> () {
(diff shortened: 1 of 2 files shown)


somiljain2006:

Sources/Markdown/Walker/Walkers/HTMLFormatter.swift
```diff
@@ -321,3 +324,14 @@ public struct HTMLFormatter: MarkupWalker {
...
 }
+
+private extension String {
+    func htmlEscaped() -> String {
+        return self
+            .replacingOccurrences(of: "&", with: "&amp;")
+            .replacingOccurrences(of: "<", with: "&lt;")
+            .replacingOccurrences(of: ">", with: "&gt;")
+            .replacingOccurrences(of: "\"", with: "&quot;")
+            .replacingOccurrences(of: "'", with: "&#39;")
```
d-ronnqvist: minor: From what I understand of the HTML specification it's only necessary to escape quotation marks and apostrophes inside the value of an attribute; for example:

```
<span id="one&qout;two"> ...
```

Sources/Markdown/Walker/Walkers/HTMLFormatter.swift
```diff
@@ -264,6 +264,9 @@ public struct HTMLFormatter: MarkupWalker {
         if let destination = link.destination {
             result += " href=\"\(destination)\""
         }
+        if let title = link.title, !title.isEmpty {
+            result += " title=\"\(title)\""
+        }
```
d-ronnqvist: I can't find where this change is verified in a test.

somiljain2006:
@d-ronnqvist Thanks for the review. I've updated the escaping helper to only escape &, <, and >, since it's used for text and code content rather than attribute values. I also removed the unrelated link title change to keep the PR focused on the HTML escaping fix, so there is no longer any untested behavior being introduced.

Tests/MarkdownTests/Visitors/HTMLFormatterTests.swift
```diff
@@ -316,4 +316,44 @@ final class HTMLFormatterTests: XCTestCase {
...
     }
+    
+    func testHtmlEscaping() {
+      do {
+        let inputText = "&lt;key&gt;"
+
+        let expectedOutput = """
+          <p>&lt;key&gt;</p>
+
+          """
```
d-ronnqvist, line 330: not-blocking but it could be useful to have a comment that briefly explains why the `&` isn't escaped into `&amp;` in this test (because the `Document` parses `&lt;` in the input and turns it into `<` in the `Text` element that the formatter visits).

d-ronnqvist (approved):
Thanks for fixing this issue.

d-ronnqvist:
@swift-ci please test

somiljain2006:
@d-ronnqvist I have added the required comment. Can you run the CI again?

d-ronnqvist:
Thanks. That comment looks great

d-ronnqvist:
@swift-ci please test

PR #276, opened 2026-06-12
Replace recursive CommonMark conversion with iterative traversal

### LINKED ISSUE #275: Document(parsing:) stack-overflows (SIGSEGV) on deeply nestable emphasis-delimiter runs
### Summary

`Document(parsing:)` crashes the process with a stack overflow (SIGSEGV) on input that produces a very deeply *nested* emphasis tree — e.g. two long runs of the same emphasis delimiter (`*` or `_`) that pair into thousands of nested `emphasis`/`strong` nodes. The crash is in swift-markdown's own Swift layer, not in the underlying C parser: the bundled `swift-cmark` is already hardened against this input class (see "Root cause" below), but swift-markdown's recursive C→Swift tree conversion and recursive `RawMarkup` deallocation have no depth bound.

Because the crash happens *inside* `Document(parsing:)` (and again later when the resulting tree is released), a caller cannot defend against it with a `do/catch` or any post-parse check — a single malicious `.md` file aborts the whole process. This is a denial-of-service vector for any app that renders untrusted Markdown.

### Reproduction

```swift
import Markdown

let run = String(repeating: "*", count: 30_000)   // "_" works identically
_ = Document(parsing: run + "x" + run)             // SIGSEGV
```

Notes on what does / doesn't trigger it:

- **Two runs that can pair are required.** A single contiguous run is safe at any size (`String(repeating: "*", count: 200_000)` parses fine — one delimiter run, nothing to match). Asymmetric runs are safe (depth ≈ `min(left, right)`).
- It is the *nesting depth* of the produced tree that matters, not byte size: many small runs in one paragraph (`String(repeating: "*x", count: 30_000)`) also crash.
- `~` (GFM strikethrough) does **not** trigger it (bounded processing).
- The exact threshold is stack-size dependent: on a default 8 MB main-thread stack it overflows around ~22,500 paired `*`; on the smaller stacks used by Swift-concurrency cooperative threads (e.g. parsing inside `Task.detached`) it overflows far lower — around ~1,400 paired `*`.

### Environment

- swift-markdown `9b43984b12577548051064599c0b426d95ec4ef8` (main)
- swift-cmark `8bab380ed057325dd5ed496fd97fd60f554d8656` (gfm)
- Swift 6.3.2, macOS 26.5 (arm64)

### Root cause (the C layer is *not* the problem)

The pathological-emphasis *input* class is known upstream in `cmark-gfm` and was addressed there as quadratic-**time** DoS bugs (e.g. GHSA-r8vr-c48j-fcc5 uses `"_"*n + "." + "_"*n`; GHSA-29g3-96g3-jg6c uses `"*t "*n + "_t*_ "*n`). Those fixes are present in the pinned `swift-cmark`:

- `process_emphasis` (`src/inlines.c`) is iterative and carries the `openers_bottom` optimization (the time fix).
- `S_free_nodes` (`src/node.c`) is **iterative** — it splices children into the sibling list and frees in a loop, so the C side does not overflow when freeing a deep tree.

So cmark itself builds a *legal but pathologically deep* node tree without crashing. The stack overflow is one layer up, in swift-markdown:

1. **Tree conversion** — `CommonMarkConverter` recursively converts the cmark C node tree into `RawMarkup`. Deep nesting → deep recursion → overflow during `Document(parsing:)`.
2. **Deallocation** — `RawMarkup` is a `ManagedBuffer` with tail-allocated children; its `deinit` releasing those children triggers each child's `deinit` in turn, so tearing down a deep tree recurses one frame per nesting level (`RawMarkup.deinit`). Even if conversion were made safe, *releasing* the returned `Document` would still overflow on the caller's stack.

(Both sites are reachable independently — I confirmed the deinit path by parsing on a large stack and then releasing the `Document` on a small stack, which still crashes.)

### Suggested fix

Bound swift-markdown's own recursion so it degrades instead of crashing:

- Make the C→Swift conversion in `CommonMarkConverter` iterative (explicit work stack), or impose a maximum nesting depth and surface it as a recoverable outcome rather than a process abort.
- Make `RawMarkup` teardown iterative (mirroring cmark's own iterative `S_free_nodes`) so deallocating a deep tree cannot overflow.

A depth cap analogous to cmark's handling of other pathological inputs would be sufficient; the goal is "never SIGSEGV on attacker-controlled input," matching the protection the C layer already has.

somiljain2006:
Bug/issue #, if applicable: Fixes #275

## Summary

Fix a stack overflow in Document(parsing:) caused by deeply nested Markdown structures.

Previously, swift-markdown converted the cmark AST to RawMarkup using recursive traversal. Inputs that produced extremely deep nesting could exhaust the call stack during conversion, causing the process to crash.

This change replaces the recursive conversion logic with an iterative implementation driven by the cmark_iter event stream. Container nodes are tracked using an explicit work stack and are finalized when their corresponding CMARK_EVENT_EXIT event is encountered. This prevents stack growth proportional to document nesting depth while preserving existing parser behavior.

Additional regression tests were added to verify:
- Deeply nested documents can be parsed without overflowing the stack.
- Multiline links continue to be parsed correctly.
- Block directive parsing preserves the structure and source locations of ordinary multiline links.
- Nested structures continue to produce the expected source ranges.

## Dependencies

N/A

## Testing

Run the full test suite

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/Markdown/Base/RawMarkup.swift
+++ b/Sources/Markdown/Base/RawMarkup.swift
@@ -162,8 +162,35 @@ final class RawMarkup: ManagedBuffer<RawMarkupHeader, RawMarkup> {
     }
 
     deinit {
-        return self.withUnsafeMutablePointerToElements {
-            $0.deinitialize(count: header.childCount)
+        var workStack = [RawMarkup]()
+        
+        self.withUnsafeMutablePointerToElements { elements in
+            for i in 0..<header.childCount {
+                workStack.append(elements[i])
+            }
+            elements.deinitialize(count: header.childCount)
+        }
+        
+        while var node = workStack.popLast() {
+            guard isKnownUniquelyReferenced(&node) else {
+                continue
+            }
+            
+            node.withUnsafeMutablePointerToElements { elements in
+                for i in 0..<node.header.childCount {
+                    workStack.append(elements[i])
+                }
+                elements.deinitialize(count: node.header.childCount)
+            }
+            
+            node.withUnsafeMutablePointerToHeader { headerPtr in
+                headerPtr.pointee = RawMarkupHeader(
+                    data: node.header.data,
+                    childCount: 0,
+                    subtreeCount: node.header.subtreeCount,
+                    parsedRange: node.header.parsedRange
+                )
+            }
         }
     }
 
--- a/Sources/Markdown/Base/RawMarkup.swift
+++ b/Sources/Markdown/Base/RawMarkup.swift
@@ -165,6 +165,12 @@ final class RawMarkup: ManagedBuffer<RawMarkupHeader, RawMarkup> {
         var workStack = [RawMarkup]()
         
         self.withUnsafeMutablePointerToElements { elements in
+            
+            // Retain children on a heap-allocated work stack before
+            // deinitializing the tail allocation. Without this, releasing a
+            // uniquely referenced child would immediately enter its deinit,
+            // recursively cascading through the tree and potentially
+            // overflowing the call stack for deeply nested markup.
             for i in 0..<header.childCount {
                 workStack.append(elements[i])
             }
--- a/Sources/Markdown/Parser/CommonMarkConverter.swift
+++ b/Sources/Markdown/Parser/CommonMarkConverter.swift
@@ -212,22 +212,6 @@ struct MarkupParser {
         return String(cString: rawText)
     }
     
-    private static func convertThematicBreak(parsedRange: SourceRange?) -> RawMarkup {
-        return .thematicBreak(parsedRange: parsedRange)
-    }
-
-    private static func convertText(node: UnsafeMutablePointer<cmark_node>!, parsedRange: SourceRange?) -> RawMarkup {
-        return .text(parsedRange: parsedRange, string: getLiteralContent(node: node))
-    }
-
-    private static func convertSoftBreak(parsedRange: SourceRange?) -> RawMarkup {
-        return .softBreak(parsedRange: parsedRange)
-    }
-
-    private static func convertLineBreak(parsedRange: SourceRange?) -> RawMarkup {
-        return .lineBreak(parsedRange: parsedRange)
-    }
-    
     /// Converts a leaf cmark node directly into its corresponding `RawMarkup` representation.
     private static func createLeaf(nodeType: CommonMarkNodeType, node: UnsafeMutablePointer<cmark_node>, parsedRange: SourceRange?, state: MarkupConverterState) -> RawMarkup {
         precondition(state.event == CMARK_EVENT_ENTER, "Expected ENTER event when creating a leaf node.")
@@ -240,13 +224,13 @@ struct MarkupParser {
         case .htmlBlock:
             return .htmlBlock(parsedRange: parsedRange, html: getLiteralContent(node: node))
         case .thematicBreak:
-            return convertThematicBreak(parsedRange: parsedRange)
+            return .thematicBreak(parsedRange: parsedRange)
         case .text:
-            return convertText(node: node, parsedRange: parsedRange)
+            return .text(parsedRange: parsedRange, string: getLiteralContent(node: node))
         case .softBreak:
-            return convertSoftBreak(parsedRange: parsedRange)
+            return .softBreak(parsedRange: parsedRange)
         case .lineBreak:
-            return convertLineBreak(parsedRange: parsedRange)
+            return .lineBreak(parsedRange: parsedRange)
         case .code:
             let literalContent = getLiteralContent(node: node)
             if state.options.contains(.parseSymbolLinks), cmark_node_get_backtick_count(node) > 1 {
--- a/Sources/Markdown/Parser/CommonMarkConverter.swift
+++ b/Sources/Markdown/Parser/CommonMarkConverter.swift
@@ -211,39 +211,105 @@ struct MarkupParser {
         }
         return String(cString: rawText)
     }
-    
+
+    private static func convertCodeBlock(_ state: MarkupConverterState) -> RawMarkup {
+        precondition(state.event == CMARK_EVENT_ENTER)
+        precondition(state.nodeType == .codeBlock)
+        let parsedRange = state.range(state.node)
+        let language = String(cString: cmark_node_get_fence_info(state.node))
+        let code = getLiteralContent(node: state.node)
+        return .codeBlock(parsedRange: parsedRange, code: code, language: language.isEmpty ? nil : language)
+    }
+
+    private static func convertHTMLBlock(_ state: MarkupConverterState) -> RawMarkup {
+        precondition(state.event == CMARK_EVENT_ENTER)
+        precondition(state.nodeType == .htmlBlock)
+        let parsedRange = state.range(state.node)
+        let html = getLiteralContent(node: state.node)
+        return .htmlBlock(parsedRange: parsedRange, html: html)
+    }
+
+    private static func convertThematicBreak(_ state: MarkupConverterState) -> RawMarkup {
+        precondition(state.event == CMARK_EVENT_ENTER)
+        precondition(state.nodeType == .thematicBreak)
+        let parsedRange = state.range(state.node)
+        return .thematicBreak(parsedRange: parsedRange)
+    }
+
+    private static func convertText(_ state: MarkupConverterState) -> RawMarkup {
+        precondition(state.event == CMARK_EVENT_ENTER)
+        precondition(state.nodeType == .text)
+        let parsedRange = state.range(state.node)
+        let string = getLiteralContent(node: state.node)
+        return .text(parsedRange: parsedRange, string: string)
+    }
+
+    private static func convertSoftBreak(_ state: MarkupConverterState) -> RawMarkup {
+        precondition(state.event == CMARK_EVENT_ENTER)
+        precondition(state.nodeType == .softBreak)
+        let parsedRange = state.range(state.node)
+        return .softBreak(parsedRange: parsedRange)
+    }
+
+    private static func convertLineBreak(_ state: MarkupConverterState) -> RawMarkup {
+        precondition(state.event == CMARK_EVENT_ENTER)
+        precondition(state.nodeType == .lineBreak)
+        let parsedRange = state.range(state.node)
+        return .lineBreak(parsedRange: parsedRange)
+    }
+
+    private static func convertInlineCode(_ state: MarkupConverterState) -> RawMarkup {
+        precondition(state.event == CMARK_EVENT_ENTER)
+        precondition(state.nodeType == .code)
+        let parsedRange = state.range(state.node)
+        let literalContent = getLiteralContent(node: state.node)
+        if state.options.contains(.parseSymbolLinks),
+           cmark_node_get_backtick_count(state.node) > 1 {
+            return .symbolLink(parsedRange: parsedRange, destination: literalContent)
+        } else {
+            return .inlineCode(parsedRange: parsedRange, code: literalContent)
+        }
+    }
+
+    private static func convertInlineHTML(_ state: MarkupConverterState) -> RawMarkup {
+        precondition(state.event == CMARK_EVENT_ENTER)
+        precondition(state.nodeType == .html)
+        let parsedRange = state.range(state.node)
+        let html = getLiteralContent(node: state.node)
+        return .inlineHTML(parsedRange: parsedRange, html: html)
+    }
+
+    private static func convertCustomInline(_ state: MarkupConverterState) -> RawMarkup {
+        precondition(state.event == CMARK_EVENT_ENTER)
+        precondition(state.nodeType == .customInline)
+        let parsedRange = state.range(state.node)
+        let text = getLiteralContent(node: state.node)
+        return .customInline(parsedRange: parsedRange, text: text)
+    }
+
     /// Converts a leaf cmark node directly into its corresponding `RawMarkup` representation.
-    private static func createLeaf(nodeType: CommonMarkNodeType, node: UnsafeMutablePointer<cmark_node>, parsedRange: SourceRange?, state: MarkupConverterState) -> RawMarkup {
-        precondition(state.event == CMARK_EVENT_ENTER, "Expected ENTER event when creating a leaf node.")
-        precondition(state.nodeType == nodeType, "MarkupConverterState nodeType does not match the leaf being created.")
-        switch nodeType {
+    private static func createLeaf(state: MarkupConverterState) -> RawMarkup {
+        switch state.nodeType {
         case .codeBlock:
-            let language = String(cString: cmark_node_get_fence_info(node))
-            let code = getLiteralContent(node: node)
-            return .codeBlock(parsedRange: parsedRange, code: code, language: language.isEmpty ? nil : language)
+            return convertCodeBlock(state)
         case .htmlBlock:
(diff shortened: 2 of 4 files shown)


somiljain2006:

d-ronnqvist:
> @d-ronnqvist Can you review this pr?

I briefly started looking at this yesterday but due to the nature of the changes it might take a while to review.

Sources/Markdown/Base/RawMarkup.swift
d-ronnqvist, line 1: Question: can you explain why it's necessary to recursively deinitialize the uniquely referenced elements like this? Is it necessary because of the iterative traversal?

I thought that deinitializing a mutable pointer would decrement the retain count of all those elements. If I add a basic `print("deinitializing \(data))")` to this deinitializer it _seems_ to confirm that.

somiljain2006: I did some testing to verify the deallocation behavior.

I added a standalone test that bypasses the parser entirely and only constructs and releases a deeply nested RawMarkup tree:

```
 func testTrulyDeepNestingDeallocation() {
      let depth = 15000

      do {
          var current = RawMarkup.text(parsedRange: nil, string: "Deep Leaf")

          for _ in 0..<depth {
              current = RawMarkup.blockQuote(parsedRange: nil, [current])
          }
      }

      XCTAssertTrue(true)
  }
```

When I run this test against the current main branch implementation of RawMarkup.deinit, it crashes during deallocation with EXC_BAD_ACCESS at the elements.deinitialize(count:) call.

image

So while elements.deinitialize(count:) does correctly release child references, there appears to be a failure mode when deallocating sufficiently deep trees using the current implementation, which the iterative work-stack approach avoids.

Sources/Markdown/Parser/CommonMarkConverter.swift
```diff
@@ -173,441 +168,180 @@ fileprivate struct MarkupConverterState {
...
-            return convertList(state)
-        case .item:
-            return convertListItem(state)
+    /// Extracts the literal text payload from a cmark node.
+    private static func getLiteralContent(node: UnsafeMutablePointer<cmark_node>!) -> String {
+        guard let rawText = cmark_node_get_literal(node) else {
+            fatalError("Expected literal content for cmark node but got null pointer")
+        }
+        return String(cString: rawText)
+    }
```
d-ronnqvist: Question: Was there a particular reason for moving this exact implementation from later in this file to here?

Sources/Markdown/Parser/CommonMarkConverter.swift
```diff
@@ -173,441 +168,180 @@ fileprivate struct MarkupConverterState {
...
+    /// Leaf nodes are converted immediately when their ENTER event is
+    /// encountered and therefore never need a `ParsingFrame`.
+    private static func isLeaf(_ type: CommonMarkNodeType) -> Bool {
+        switch type {
+        case .codeBlock, .htmlBlock, .thematicBreak, .text, .softBreak, .lineBreak, .code, .html, .customInline:
+            return true
+        default:
+            return false
+        }
+    }
```
d-ronnqvist: minor: It would likely be more idiomatic to define this as a private extension on `CommonMarkNodeType` that adds a computed property that switches over self.

Sources/Markdown/Parser/CommonMarkConverter.swift
```diff
@@ -173,441 +168,180 @@ fileprivate struct MarkupConverterState {
...
         case .taskListItem:
-            return convertTaskListItem(state)
+            let checkbox: Checkbox = cmark_gfm_extensions_get_tasklist_item_checked(node) ? .checked : .unchecked
+            return .listItem(checkbox: checkbox, parsedRange: parsedRange, children)
         case .table:
-            return convertTable(state)
+            let columnCount = Int(cmark_gfm_extensions_get_table_columns(node))
+            let columnAlignments = (0..<columnCount).map { column -> Table.ColumnAlignment? in
+                let ascii = cmark_gfm_extensions_get_table_alignments(node)[column]
+                switch Character(UnicodeScalar(ascii)) {
```
d-ronnqvist: nit: I don't think it' necessary to create a full `Character` for this comparison. I'm pretty sure that the `UnicodeScalar` would suffice for the pattern matching below.

Sources/Markdown/Parser/CommonMarkConverter.swift
```diff
@@ -173,441 +168,180 @@ fileprivate struct MarkupConverterState {
...
-        let parsedRange = state.range(state.node)
-        let childConversion = convertChildren(state)
-        precondition(childConversion.state.node == state.node)
-        precondition(childConversion.state.event == CMARK_EVENT_EXIT)
-        return MarkupConversion(state: childConversion.state.next(), result: .listItem(checkbox: .none, parsedRange: parsedRange, childConversion.result))
-    }
-
-    private static func convertCodeBlock(_ state: MarkupConverterState) -> MarkupConversion<RawMarkup> {
-        precondition(state.event == CMARK_EVENT_ENTER)
-        precondition(state.nodeType == .codeBlock)
```
d-ronnqvist, line 324: I wonder if It would make sense to keep all these preconditions to ensure that the implementation is in the right state when converting each type of element.

Sources/Markdown/Parser/CommonMarkConverter.swift
```diff
@@ -173,441 +168,180 @@ fileprivate struct MarkupConverterState {
...
+            return .thematicBreak(parsedRange: parsedRange)
         case .text:
-            return convertText(state)
+            return .text(parsedRange: parsedRange, string: getLiteralContent(node: node))
         case .softBreak:
-            return convertSoftBreak(state)
+            return .softBreak(parsedRange: parsedRange)
         case .lineBreak:
-            return convertLineBreak(state)
+            return .lineBreak(parsedRange: parsedRange)
```
d-ronnqvist: minor: It would be easier to review this PR if it didn't move so much code around. For these lines specifically; it has inlined `convertThematicBreak`, `convertText`, `convertSoftBreak`, and `convertLineBreak` without the preconditions.

If instead that code had remained in those private functions and their return type had changed from `MarkupConversion<RawMarkup>` to `RawMarkup` then it would have been much easier to compare the before and after and know what changed.

Sources/Markdown/Parser/CommonMarkConverter.swift
```diff
@@ -173,441 +168,180 @@ fileprivate struct MarkupConverterState {
...
+                header = .tableHead(parsedRange: nil, columns: [])
+            }
+
+            let body: RawMarkup
+            if !mutableChildren.isEmpty {
+                let pendingBodyRange = state.pendingTableBody?.range
+                body = .tableBody(parsedRange: pendingBodyRange, rows: mutableChildren)
+            } else {
+                body = .tableBody(parsedRange: nil, rows: [])
+            }
```
d-ronnqvist: minor: There used to be a `precondition` that `state.pendingTableBody == nil` when `children.isEmpty`. If that precondition is still true, then this if-statement isn't necessary.
```suggestion
            if mutableChildren.isEmpty {
                precondition(state.pendingTableBody == nil)
            }
            let body = RawMarkup.tableBody(parsedRange: pendingBodyRange, rows: mutableChildren)
```

Sources/Markdown/Base/RawMarkup.swift
somiljain2006: I dug into the crash a bit further using the debugger. It is correct that elements.deinitialize(count:) releases the child references. The issue seems to be what happens after those releases.

Looking at the failing stack trace from the standalone deallocation test, the pattern repeats this many times:

RawMarkup.deinit
_swift_release_dealloc
swift_arrayDestroy
UnsafeMutablePointer.deinitialize
...

Each deinitialize(count:) call releases a child, that child immediately enters RawMarkup.deinit, which then calls deinitialize(count:) on its own children, and so on. The destruction path becomes recursive with depth proportional to the tree depth.

The standalone test consistently crashes on main once the nesting becomes sufficiently deep (around 10k levels), whereas the iterative implementation completes successfully. That seems to indicate the failure mode is the recursive ARC teardown during deallocation rather than the parsing/construction phase.

d-ronnqvist (commented):
This code removes a lot of preconditions that verify that the implementation is in the right state at each point of the conversion. For my taste it would be good to add those back unless @QuietMisdreavus feels otherwise.

Sources/Markdown/Base/RawMarkup.swift
d-ronnqvist: Thanks for explaining that to me and thanks for digging deeper into the issue. IMO It would be useful to phrase some of that explanation into a code comment inside the `deinit` implementation. Otherwise, someone in the future might look at it and wonder (like I did) if it's really necessary.

Sources/Markdown/Parser/CommonMarkConverter.swift
somiljain2006: No particular reason beyond readability. During the refactor I grouped the helper methods (getLiteralContent, isLeaf, createLeaf, createContainer) together near the top of MarkupParser because they're all used by the iterative conversion logic. I'm okay to move it back if you think the previous location reads better.

Sources/Markdown/Parser/CommonMarkConverter.swift
```diff
@@ -246,368 +211,158 @@ struct MarkupParser {
...
-        var layout = [RawMarkup]()
-
-        while state.node != root && state.event != CMARK_EVENT_EXIT {
-            let conversion = convertAnyElement(state)
-            layout.append(conversion.result)
-            state = conversion.state
-        }
-        return MarkupConversion(state: state, result: layout)
+    
+    private static func convertThematicBreak(parsedRange: SourceRange?) -> RawMarkup {
```
d-ronnqvist: minor: It would be easier to review this if it didn't move from it's previous location on [lines 371-376](https://github.com/swiftlang/swift-markdown/pull/276/changes#diff-775d3752fc00ba17abc7a3355dbae53b50643f9a62d645acebd7afecab6731b7L371-L376).

This applies to all of the moved code.

somiljain2006:
@d-ronnqvist Is the code readability better now?

d-ronnqvist:
> @d-ronnqvist Is the code readability better now?

Yes, thank you. It's much easier to review the changes now.

Sources/Markdown/Parser/CommonMarkConverter.swift
```diff
@@ -626,28 +431,66 @@ struct MarkupParser {
...
+                    precondition(!stack.isEmpty, "Leaf node encountered without a parent document on the stack.")
+                    stack[stack.count - 1].children.append(leaf)
+                } else {
+                    stack.append(ParsingFrame(node: node, nodeType: nodeType, parsedRange: parsedRange))
+                }
+                state = state.next()
+                
+            } else if state.event == CMARK_EVENT_EXIT {
+                if nodeType.isLeaf {
+                    state = state.next()
```
d-ronnqvist: minor: Based on [this comment](https://github.com/swiftlang/swift-cmark/blob/gfm/src/include/cmark-gfm.h#L274-L284) in the cmark headers, this code path should seemingly never happen.

> Iterators will never return `EXIT` events for leaf nodes, which are nodes of type:
>
> * `CMARK_NODE_HTML_BLOCK`
> * `CMARK_NODE_THEMATIC_BREAK`
> * `CMARK_NODE_CODE_BLOCK`
> * `CMARK_NODE_TEXT`
> * `CMARK_NODE_SOFTBREAK`
> * `CMARK_NODE_LINEBREAK`
> * `CMARK_NODE_CODE`
> * `CMARK_NODE_HTML_INLINE`

That would mean that the `isLeaf` check could be covered by a precondition, giving the body of the else-branch one level less of indentation.

Sources/Markdown/Parser/CommonMarkConverter.swift
```diff
@@ -626,28 +431,66 @@ struct MarkupParser {
...
+                        stack[stack.count - 1].children.append(container)
+                        if nodeType == .table {
+                            state = state.next(clearPendingTableBody: true)
+                        } else {
+                            state = state.next()
+                        }
+                    }
+                }
+            } else {
+                state = state.next()
```
d-ronnqvist: minor: There are only 4 events; `NONE`, `DONE`, `ENTER`, and `EXIT`. Because both `ENTER` and `EXIT` are covered by the if-statements above, this else-branch only (theoretically) covers the `NONE` and `DONE` events (although [this documentation](https://github.com/swiftlang/swift-cmark/blob/gfm/src/include/cmark-gfm.h#L310-L311) in cmark would indicate that `cmark_iter_next` doesn't return a `NONE` event—which [the current implementation](https://github.com/swiftlang/swift-cmark/blob/gfm/src/iterator.c#L42-L79) confirms—so in practice this else-branch only covers the `DONE` event). 

Calling next on a `DONE` event [returns `DONE` again](https://github.com/swiftlang/swift-cmark/blob/gfm/src/iterator.c#L49-L51), so I think this else-branch can safely be removed.

Sources/Markdown/Parser/CommonMarkConverter.swift
```diff
@@ -626,28 +431,66 @@ struct MarkupParser {
...
+        precondition(state.event == CMARK_EVENT_ENTER)
+        precondition(state.nodeType == .document)
 
-        precondition(initialState.node != nil)
+        var stack = [ParsingFrame]()
 
-        cmark_node_free(initialState.node)
-        cmark_iter_free(finalState.iterator)
-        cmark_parser_free(parser)
+        while state.event != CMARK_EVENT_DONE && state.event != CMARK_EVENT_NONE {
```
d-ronnqvist: minor: As far as I can tell from the cmark documentation and implementation, the event never returns to `NONE`. Because the precondition above verifies that the start event is `ENTER` and because `cmark_iter_next` is [documented](https://github.com/swiftlang/swift-cmark/blob/gfm/src/include/cmark-gfm.h#L310-L311) to only return `DONE`, `ENTER`, or `EXIT` we should be safe to only check `state.event != CMARK_EVENT_DONE` in this while-loop.

Sources/Markdown/Parser/CommonMarkConverter.swift
```diff
@@ -626,28 +431,66 @@ struct MarkupParser {
...
+                        let data = _MarkupData(AbsoluteRawMarkup(markup: container, metadata: MarkupMetadata(id: .newRoot(), indexInParent: 0)))
+                        return makeMarkup(data) as! Document
+                        
+                    } else {
+                        stack[stack.count - 1].children.append(container)
+                        if nodeType == .table {
+                            state = state.next(clearPendingTableBody: true)
+                        } else {
+                            state = state.next()
+                        }
```
d-ronnqvist: minor (non-blocking): It could perhaps be easier for the reader to see that both branches call the same  function if the boolean parameter was passed inline:

```suggestion
                        state = state.next(clearPendingTableBody: nodeType == .table)
```

d-ronnqvist (commented):
Thanks for both opening this PR and for bearing with me, making all these changes to make the differences easier to review. 

I've looked at all the code now and suggested a few places in the main iteration loop where handling of unexpected states could either be removed or changed into a precondition to ever-so-slightly simplify the happy path of the loop. Other than those minor remarks, the code looks really good to me.

somiljain2006:
@d-ronnqvist Thanks for the review. I have addressed all the minor remarks. Can you review them?

d-ronnqvist (approved):
Thanks for fixing this issue. This looks really good to me.

d-ronnqvist:
@swift-ci please test

PR #281, opened 2026-07-02
CMake: Only emit the build-tree export for a top-level build

rintaro:
The build-tree `export(TARGETS …)` in `cmake/modules/CMakeLists.txt` is only useful for the standalone install-and-`find_package` flow, where cmark-gfm is an imported package whose targets need no build-tree export set. When SwiftMarkdown is built as a subproject (e.g. via `FetchContent`), the consumer references its targets directly and never reads the generated export file, so the export serves no purpose there — and it actively fails, because the exported interface pulls in cmark-gfm's locally-built targets such as `libcmark-gfm-extensions`, which aren't in this build tree's export set:

```
export called with target "Markdown" which requires target
"libcmark-gfm-extensions" that is not in any export set.
```

Restrict the `export()` to top-level builds. This is behavior-preserving for the standalone build (the export runs exactly as before) and unblocks subproject consumers such as swift-format.

The guard uses `PROJECT_NAME STREQUAL CMAKE_PROJECT_NAME` rather than `PROJECT_IS_TOP_LEVEL`, which requires CMake 3.21 while this project's minimum is 3.19.

(This should unblock https://github.com/swiftlang/swift-format/pull/1241)


rintaro:
@swift-ci test

QuietMisdreavus (approved):

rintaro:
@swift-ci test

PR #282, opened 2026-07-03
Fix inline code wrapping idempotence

### LINKED ISSUE #197: MarkupFormatter is inserting extra newlines
I've noticed that the formatting has been putting in extra newlines when wrapping text that contains inline code blocks. I've included a test below that illustrates the problem; both tests fail, but in particular the first fails because there is an extra newline preceding `contributesBlockIndent` in the output:

```
fileprivate func markdownFormat(_ text: String, _ maxLength: Int) -> String {
  let document = Document(parsing: text, options: .disableSmartOpts)
  let lineLimit = MarkupFormatter.Options.PreferredLineLimit(
    maxLength: maxLength,
    breakWith: .softBreak
  )
  let formatterOptions = MarkupFormatter.Options(
    orderedListNumerals: .incrementing(start: 1),
    useCodeFence: .always,
    condenseAutolinks: false,
    preferredLineLimit: lineLimit
  )

  let output = document.format(options: formatterOptions)
  return output
}

@Test func testInlineCodeWrapping() {
  let str = """
    This indent is applied independently of `contributesBlockIndent`, which means a given break \
    may apply both a continuation indent and a block indent, either indent, or neither indent.
    """

  let expected = """
    This indent is applied independently of 
    `contributesBlockIndent`, which means a given break
    may apply both a continuation indent and a block
    indent, either indent, or neither indent.
    """

  // format it twice; we want (a) the first version to be correct, and (b) the second version to be the same
  let trip1 = markdownFormat(str, 55)
  let trip2 = markdownFormat(trip1, 55)

  #expect(trip1 == expected)
  #expect(trip1 == trip2)
}
```

somiljain2006:
Bug/issue #, if applicable: Fixes #197

## Summary

Fix inline code wrapping idempotence and trailing whitespace handling

Prevent extra newlines and trailing spaces when wrapping inline code by trimming trailing whitespace before queued line breaks and preserving formatter state during speculative inline code layout. This restores idempotency across repeated formatting passes.

## Dependencies

N/A

## Testing

Adds regression tests that cover inline code wrapping in paragraphs, lists, ordered lists, blockquotes, and existing soft-break scenarios

Steps:
Run MarkupFormatterTests

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/Markdown/Walker/Walkers/MarkupFormatter.swift
+++ b/Sources/Markdown/Walker/Walkers/MarkupFormatter.swift
@@ -579,6 +579,19 @@ public struct MarkupFormatter: MarkupWalker {
         state.newlineStreak = 0
     }
 
+    /// Remove trailing spaces that have already been printed on the current line.
+    private static func trimTrailingSpaces(
+        from currentResult: inout String,
+        state currentState: inout State,
+        preserving minimumLineLength: Int = 0
+    ) {
+        while currentResult.last == " " && currentState.lastLineLength > minimumLineLength {
+            currentResult.removeLast()
+            currentState.currentLength -= 1
+            currentState.lastLineLength -= 1
+        }
+    }
+
     /// Print raw text while visiting an element, wrapping automatically with
     /// soft or hard line breaks.
     ///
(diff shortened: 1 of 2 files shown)


somiljain2006:

PR #292, opened 2026-09-08
cmake: recognize FreeBSD powerpc64le/powerpc64 in get_swift_host_arch

pkubaj:
## Summary

_Provide a description of what your PR addresses, explaining the expected user experience. 
get_swift_host_arch() maps CMAKE_SYSTEM_PROCESSOR to a Swift architecture
name using the Linux spellings (ppc64le, ppc64).  On FreeBSD the
processor is reported as powerpc64le / powerpc64, so the PowerPC host was
falling through to the "Unrecognized architecture" error.  Accept the
FreeBSD spellings alongside the existing ones.

## Testing

Build on FreeBSD / powerpc64le machine.

## Checklist

- [ ] Added tests - doesn't require new tests - FreeBSD is already support, ppc64le as well.
- [ ] Updated documentation if necessary - not necessary


QuietMisdreavus (approved):
Thanks!

